PR #18, opened 2017-11-07
fix: add closing bracket after parameter name

chmyga:
### Requirements


chmyga:
Pushed by mistake.

PR #25, opened 2017-11-09
feature(TCOMP-650):Test component in beam pipeline

chmyga:
* Add test to run component in beam pipeline

### Requirements

### Why this PR is needed?
To ensure component can be executed in beam pipelines
Test to run components in beam pipeline


rmannibucau:
We already have these tests into the project in beam module. As checked together offline, we must just have a manual validation test phase on a "big" spark cluster (bigger than the one we can tolerate in the upstream suite) so I will close this PR for now.

PR #26, opened 2017-11-09
Igonchar/tcomp 680

gonchar-ivan:
### Requirements

It implements ComponentModel methods, which allows to see component on design canvas


component-design-manager/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>component-runtime</artifactId>
+    <groupId>org.talend.sdk.component</groupId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>component-design-manager</artifactId>
```
rmannibucau: Id go with component-runtime-design-extension (and corresponding name) since it doesnt have a manager by itself

component-design-manager/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
```
rmannibucau: Drop if useless ;)

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignModel.java
```diff
@@ -0,0 +1,47 @@
...
+@Data
+public class DesignModel {
+
+    /**
+     * Component unique id
+     */
+    private final String id;
+    
+    /**
+     * Component name
```
rmannibucau: If the comment is useless no need to put it ;)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentFamilyMeta.java
```diff
@@ -126,6 +135,26 @@ public ComponentBundle findBundle(final ClassLoader loader, final Locale locale)
...
             });
         }
+        
+        /**
+         * Returns a {@link Collection} of input flows names of this Component
+         * Should be overridden in inheritors
+         * 
+         * @return input flows names collection
+         */
+        public Collection<String> getInputFlows() {
```
rmannibucau: Abstract if 100% dependent from the child - no loose default ;)

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/ReturnVariables.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.talend.sdk.component.studio.model;
+
+/**
+ * Stores return variables constants
+ */
+public final class ReturnVariables {
```
rmannibucau: NoArgConstructor(access=PRIVATE) or move to an enum

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignModel.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    private final String name;
```
rmannibucau: We dont need the name if we have the id

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignContainerListener.java
```diff
@@ -0,0 +1,68 @@
...
+     * It depends on Updater listener which adds {@link ContainerComponentRegistry} class to {@link Container}
+     */
+    @Override
+    public void onCreate(Container container) {
+        ContainerComponentRegistry componentRegistry = container.get(ContainerComponentRegistry.class);
+        if (componentRegistry == null) {
+            throw new IllegalArgumentException("container doesn't contain ContainerComponentRegistry");
+        }
+
+        DesignModelRegistry models = new DesignModelRegistry();
```
rmannibucau: Why not add a map in basemeta for this kind of extension and avoid to multiply the registries? Then we just store DesignModel and no join to do

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentFamilyMeta.java
```diff
@@ -126,6 +135,26 @@ public ComponentBundle findBundle(final ClassLoader loader, final Locale locale)
                 }
             });
         }
+
```
rmannibucau: This code was not modified to use the new model right? (The mojo as well btw)

component-design-manager/pom.xml
gonchar-ivan: Do you mean "change artifactId and name to component-runtime-design-extension"?
(Initially, I didn't understand where it should go and which id? lol)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentFamilyMeta.java
gonchar-ivan: It wasn't. Work is in progress. Things to do:
1. Move getInput/OutputFlows methods from Meta to DesignModel (or Listener)
2. Change component-server to use it
3. Change mojo to use it
4. Check in Studio

component-design-manager/pom.xml
rmannibucau: yes for the artifactId and Component Runtime :: Design Extension for the name?

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignContainerListener.java
gonchar-ivan: map<Class<?>, Object> data?
Similar as in Container?

rmannibucau: Yep, same pattern but contextual to a single component

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignModel.java
gonchar-ivan: As I understand ComponentManager is to be used for runtime, so we don't want have there things not related in runtime. 
My idea is to move eventually name and icon from BaseMeta to DesignModel as these info is not needed in runtime, but needed in design.

rmannibucau: we need the name into the runtime since it can be used for lookups (id is really just a workaround to pass it through the URL). +1 to move icons

component-design-manager/pom.xml
gonchar-ivan: done

component-design-manager/pom.xml
gonchar-ivan: done

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignContainerListener.java
gonchar-ivan: done

component-design-manager/src/main/java/org/talend/sdk/component/design/manager/DesignModel.java
gonchar-ivan: removed name from Model. 
icons to be done

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentFamilyMeta.java
gonchar-ivan: Moved from BaseMeta to FlowsFactory class. Made it abstract in parent class

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/ReturnVariables.java
gonchar-ivan: done

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -386,9 +388,16 @@ public ComponentManager(final File m2, final String dependenciesResource, final
...
+        toStream(loadServiceProviders(ContainerListenerExtension.class)).forEach(listener -> container.registerListener(listener));  
+        this.extensions =  toStream(loadServiceProviders(ComponentExtension.class)).collect(toList());
+    }
+    
+    private static <T> Stream<T> toStream(Iterator<T> iterator) {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator, Spliterator.IMMUTABLE), false);
+    }
+    
+    private static <T> Iterator<T> loadServiceProviders(Class<T> service) {
+        return ServiceLoader.load(service, Thread.currentThread().getContextClassLoader()).iterator();
```
rmannibucau: the classloader is defined in org.talend.sdk.component.runtime.manager.ComponentManager#ComponentManager, can you please ensure we use a single instance and don't rely on the TCCL in multiple places in case we refactor the code later on? (just define a ClassLoader var in the constructor and pass it to the loadServiceProviders)

PR #27, opened 2017-11-13
feature(TCOMP-650): enrich documentation

chmyga:
### Requirements

### DIFF
--- a/documentation/src/main/jbake/content/documentation.adoc
+++ b/documentation/src/main/jbake/content/documentation.adoc
@@ -1096,11 +1096,11 @@ public class Main {
         Pipeline pipeline = Pipeline.create();
         //Create beam input from mapper and apply input to pipeline
         pipeline.apply(TalendIO.read(manager.findMapper(manager.findMapper("sample", "reader", 1, new HashMap<String, String>() {{
-                    put("file", "/tmp/input.csv");
+                    put("fileprefix", "input");
                 }}).get()))
         //Create beam processor from talend processor and apply to pipeline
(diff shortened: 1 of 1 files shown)


documentation/src/main/jbake/content/documentation.adoc
```diff
@@ -16,6 +16,9 @@ Talend Component framework relies on several primitive components.
 They can all use `@PostConstruct` and `@PreDestroy` to initialize/release
 some underlying resource at the beginning/end of the processing.
 
+IMPORTANT: in distributed environments class' constructor will be called on driver node, methods annotated with
```
rmannibucau: driver/worker is not obvious and only relevant in spark context, maybe adding a schema (you can use asciidoctor-diagram) to represent the workflow (partition plan computation then execution) can help and put a name on each phases, wdyt?

documentation/src/main/jbake/content/documentation.adoc
```diff
@@ -1078,10 +1088,44 @@ For beam case, you need to rely on beam pipeline definition and use `component-r
...
 
+[source,java]
+----
+public class Main {
+    public static void main(final String[] args) {
+        final ComponentManager manager = ComponentManager.instance()
+        Pipeline pipeline = Pipeline.create();
+        //Create beam input from mapper and apply input to pipeline
+        pipeline.apply(TalendIO.read(manager.findMapper(manager.findMapper("sample", "reader", 1, new HashMap<String, String>() {{
+                    put("file", "/tmp/input.csv");
```
rmannibucau: we shouldn't hardcode this kind of path because it doesn't work this way in the big data env, it is more a pattern so maybe rename "file", "fileprefix" and remove the .csv from the value?

PR #28, opened 2017-11-15
Igonchar/tcomp 680 icon

gonchar-ivan:
### Requirements

### Why this PR is needed?

For better architecture

Icon is moved from ComponentFamilyMeta class (which is intended to use at runtime only) to DesignModel class (which is intended to use at design)


rmannibucau:
will be done later, closing for now

PR #49, opened 2017-11-24
auto formatting the whole source base

rmannibucau:
auto formatting the whole source base


sgandon:
Why don't you use the formatter proposed in the tools repository ?
If you are not happy with it then please feel free to contribute.

rmannibucau:
Cause I took another one I had locally which was working with java 8. Will update it and keep it if the last version is j8 friendly. Thanks for the heads up.

PR #107, opened 2017-12-27
feat(TCOMP-683): add schema

chmyga:
* Add schema handling to runtime

### Requirements

### DIFF
--- a/component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SchemaSettingsCreator.java
+++ b/component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SchemaSettingsCreator.java
@@ -15,6 +15,8 @@
  */
 package org.talend.sdk.component.studio.model.parameter;
 
+import java.util.List;
+
 import org.talend.core.model.process.EComponentCategory;
 import org.talend.core.model.process.EConnectionType;
 import org.talend.core.model.process.EParameterFieldType;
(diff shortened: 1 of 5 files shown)


component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SchemaSettingsCreator.java
```diff
@@ -30,17 +32,30 @@
...
      * @param category Parameter category
      * @param redrawParameter Parameter which defines whether UI should be redrawn
      */
     public SchemaSettingsCreator(final IElement iNode, final EComponentCategory category,
             final ElementParameter redrawParameter) {
         super(iNode, category, redrawParameter);
-        if (category == EComponentCategory.BASIC) {
+    }
+
+    public List<ElementParameter> getSettings() {
```
gonchar-ivan: Please, add @Override annotation

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/SchemaSettingsCreator.java
```diff
@@ -30,17 +32,30 @@
...
+        boolean hasSchema = false;
+
+        for (TaCoKitElementParameter param : settings.values()) {
+            if (param.getFieldType() == EParameterFieldType.SCHEMA_TYPE) {
+                hasSchema = true;
+                break;
+            }
+        }
+        if (!hasSchema) {
```
gonchar-ivan: please, extract  `createSchemaParameter(String schemaName)` method and move it to parent class to avoid duplicated code

PR #126, opened 2018-01-16
fix add schema widget

akhabali:
### Requirements


cmeng-talend:
LGTM

PR #140, opened 2018-01-24
fix: fix reject flow

chmyga:
* Add reject flow to connectors list

* Add schemas for all connectors, which need this

* Fix generic in javajet

### Requirements

### DIFF
--- a/component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
+++ b/component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
@@ -238,7 +238,7 @@ public class ComponentModel extends AbstractBasicComponent {
     @Override // TODO This is dummy implementation. Correct impl should be added soon
     public List<? extends IElementParameter> createElementParameters(final INode node) {
         ElementParameterCreator creator = new ElementParameterCreator(this, detail, node);
-        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters(node);
+        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters();
         return parameters;
     }
 
--- a/component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
+++ b/component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
@@ -238,7 +238,7 @@ public class ComponentModel extends AbstractBasicComponent {
     @Override // TODO This is dummy implementation. Correct impl should be added soon
     public List<? extends IElementParameter> createElementParameters(final INode node) {
         ElementParameterCreator creator = new ElementParameterCreator(this, detail, node);
-        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters();
+        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters(node);
         return parameters;
     }
 
--- a/component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
+++ b/component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
@@ -21,6 +21,7 @@ import static org.talend.core.model.process.EComponentCategory.BASIC;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.talend.core.CorePlugin;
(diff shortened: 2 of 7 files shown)


component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
```diff
@@ -238,7 +238,7 @@ public String getVersion() {
     @Override // TODO This is dummy implementation. Correct impl should be added soon
     public List<? extends IElementParameter> createElementParameters(final INode node) {
         ElementParameterCreator creator = new ElementParameterCreator(this, detail, node);
-        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters();
+        List<IElementParameter> parameters = (List<IElementParameter>) creator.createParameters(node);
```
gonchar-ivan: seems, node parameter is not required here. It is passed to ElementParameterCreator constructor.
Thus, you don't need to pass it to ElementParameterCreator.addComponentParameters() method. You may use the field directly

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
```diff
@@ -100,12 +101,36 @@ private void addComponentParameters() {
...
+    }
 
+    /**
+     * Check whether all required schema settings are created and create ones we still need depending on connectors we
+     * have.
+     * 
+     * @param node
+     * @param mainSettingsCreator
+     */
+    protected void checkSchemaProperties(final INode node, final SettingsCreator mainSettingsCreator) {
```
gonchar-ivan: seems, `node` parameter is not required here. It is already available as a class field

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
```diff
@@ -100,12 +101,36 @@ private void addComponentParameters() {
...
+        // Get all schema parameters created for current component
+        final Set<String> schemasPresent = parameters
+                .stream()
+                .filter(p -> EParameterFieldType.SCHEMA_TYPE.equals(p.getFieldType()))
+                .map(p -> p.getContext())
```
gonchar-ivan: seems, you may use method reference here:
```
IElementParameter::getContext()
```

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
```diff
@@ -100,12 +101,36 @@ private void addComponentParameters() {
...
+                .collect(Collectors.toSet());
+        // Get all connectors without schema parameter for them
+        final Set<String> connectorNames = component
+                .createConnectors(node)
+                .stream()
+                .filter(c -> c.getDefaultConnectionType().hasConnectionCategory(IConnectionCategory.FLOW)
+                        && !schemasPresent.contains(c.getName()))
+                .map(c -> c.getName())
```
gonchar-ivan: method reference may be used here

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
```diff
@@ -100,12 +101,36 @@ private void addComponentParameters() {
...
+                .collect(Collectors.toSet());
+        // Create schema parameter for each connector without schema parameter
+        for (String connectorWithoutSchema : connectorNames) {
+            parameters.add(mainSettingsCreator.createSchemaParameter(connectorWithoutSchema,
```
gonchar-ivan: just a note:
This loop may also be rewritten in following way.
```
connectorNames.forEach(c -> parameters.add(mainSettingsCreator.createSchemaParameter(c, "SCHEMA_" + c)));
```
However, your current code looks clearer for me than a code with "forEach".

component-studio-integration/src/main/java/org/talend/sdk/component/studio/ComponentModel.java
chmyga: That's reasonable! thanks! I missed that. I'll fix it.

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
chmyga: I would leave that as is, if you don't mind.

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
chmyga: fixed.

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
chmyga: fixed

component-studio-integration/src/main/java/org/talend/sdk/component/studio/model/parameter/ElementParameterCreator.java
chmyga: fixed

chmyga:
Do not merge yet, please. Found a problem with this PR.

PR #157, opened 2018-02-01
fix build element meta - ignore services

akhabali:
### Requirements


rmannibucau:
Think we must drop the services keyset since ut would break all tooling which doesnt rely on scanning (doc generation, validator, ...)

PR #182, opened 2018-02-19
add a job builder and job execution impl

akhabali:
### Requirements


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/internal/JobImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+
+    public static class InputRunner {
+
+        private final Input input;
+
+        public InputRunner(final Mapper mapper) {
+            final long totalSize = mapper.assess();
+            mapper.start();
```
rmannibucau: should be done before assess

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/internal/JobImpl.java
```diff
@@ -0,0 +1,427 @@
...
+            final Iterator<Mapper> iterator;
+            try {
+                iterator = mapper.split(totalSize).iterator();
+            } finally {
+                mapper.stop();
+            }
+            final ChainedMapper chainedMapper = new ChainedMapper(mapper, iterator);
```
rmannibucau: misses stop call?

PR #205, opened 2018-03-01
ignore null query params

akhabali:
### Requirements


rmannibucau:
Empty can be fine and just be handled as ?foo. Maybe we need an attribute in Query to say skip if empty. That said null should be enough and empty would just skip the = in the concatenation, no?

PR #208, opened 2018-03-05
change error message when file is not found

akhabali:
### Requirements


rmannibucau:
Would be great to have the actual cause of the exception and not only a partial cause, both missing file and wrong archive type are possible.

PR #259, opened 2018-06-29
auto-clean this module to get a correct shade

akhabali:
### Requirements


component-runtime-testing/component-runtime-http-junit/pom.xml
```diff
@@ -115,23 +115,13 @@
         <artifactId>maven-clean-plugin</artifactId>
         <executions>
           <execution>
-            <id>forced-clean</id>
```
rmannibucau: No, we don't awnt to execute clean at initialize since it will impact all commands, we just want to do it just before the packaging if the jar is already here. Master should have fixed the config

PR #262, opened 2018-07-02
chore: fix webapp submit

jmfrancois:
### Requirements

### Why this PR is needed?

the submit is broken

the submit now display the results

screen shot 2018-07-02 at 17 36 50

### DIFF
--- a/component-tools-webapp/src/main/frontend/src/components/Detail/Detail.component.js
+++ b/component-tools-webapp/src/main/frontend/src/components/Detail/Detail.component.js
@@ -16,6 +16,7 @@
 
 import React from 'react';
 import { Inject } from '@talend/react-cmf';
+import kit from 'component-kit.js';
 
 function NoSelectedComponent() {
 	return (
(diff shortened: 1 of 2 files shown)


component-tools-webapp/src/main/frontend/src/components/Detail/Detail.component.js
```diff
@@ -16,6 +16,7 @@
 
 import React from 'react';
 import { Inject } from '@talend/react-cmf';
+import kit from 'component-kit.js';
```
jmfrancois: add missing import

component-tools-webapp/src/main/frontend/src/components/Detail/Detail.connect.js
```diff
@@ -29,6 +29,7 @@ import Detail from './Detail.component';
 function mapStateToProps(state) {
 	return {
 		definitionURL: get(state, 'app.componentsList.selectedNode.$$detail'),
+		...state.app.component,
```
jmfrancois: add { submited, properties } from the store

PR #308, opened 2018-09-12
TCOMP-1051 : Remove brave and move to geronimo-opentracing

akhabali:
### Requirements


component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/OpenTracingConfigSource.java
```diff
@@ -19,14 +19,31 @@
...
+
     private Map<String, String> configuration = new HashMap<String, String>() {
 
         {
-            final String isTracingOn = String
-                    .valueOf(Boolean.getBoolean("TRACING_ON") && Integer.getInteger("TRACING_SAMPLING_RATATE", 0) != 0);
+            final boolean tracingOn =
+                    System.getenv(TRACING_ON) != null && Boolean.parseBoolean(System.getenv(TRACING_ON));
+            int tracingRate = 1;
+            if (System.getenv(TRACING_SAMPLING_RATE) != null) {
```
rmannibucau: either use ofNullable or at least a variable to not access twice the env please

component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/OpenTracingConfigSource.java
```diff
@@ -19,14 +19,31 @@
...
+
+    private final String TRACING_SAMPLING_RATE = "TRACING_SAMPLING_RATE";
+
     private Map<String, String> configuration = new HashMap<String, String>() {
 
         {
-            final String isTracingOn = String
-                    .valueOf(Boolean.getBoolean("TRACING_ON") && Integer.getInteger("TRACING_SAMPLING_RATATE", 0) != 0);
+            final boolean tracingOn =
+                    System.getenv(TRACING_ON) != null && Boolean.parseBoolean(System.getenv(TRACING_ON));
```
rmannibucau: you can parse directly, null will return false

PR #315, opened 2018-09-19
TCOMP-1074 : rise an error while deploying a conflicting component ve…

akhabali:
…rsion to the studio and add an option to enforce the deployment if needed

### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-1074

### DIFF
--- a/component-tools/src/main/java/org/talend/sdk/component/tools/StudioInstaller.java
+++ b/component-tools/src/main/java/org/talend/sdk/component/tools/StudioInstaller.java
@@ -75,7 +75,7 @@ public class StudioInstaller implements Runnable {
         final String artifact = mvnMeta[1];
         final String version = mvnMeta[2];
 
-        // -1. check if component can be installed.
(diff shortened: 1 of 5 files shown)


component-tools/src/main/java/org/talend/sdk/component/tools/StudioInstaller.java
```diff
@@ -68,6 +71,38 @@ public void run() {
...
+            }
+        }
+        String registry = config.getProperty("component.java.registry",
+                config.getProperty("talend.component.server.component.registry"));
+        if (!enforceDeployment && registry != null && new File(registry).exists()) {
+            final Properties components = new Properties();
+            try (final Reader reader = new FileReader(registry)) {
+                components.load(reader);
+                if (components.containsKey(artifact)) {
+                    final String installedVersion = components.get(artifact).toString().split(":")[2];
```
rmannibucau: getProperty, avoids the toString()

PR #330, opened 2018-10-29
fix(TCOMP-1129): fix of resolve $selfReference for ActionService

bchen-talend:
### Requirements

### DIFF
--- a/component-form/component-server-proxy-parent/component-server-proxy/src/main/java/org/talend/sdk/component/proxy/service/ActionService.java
+++ b/component-form/component-server-proxy-parent/component-server-proxy/src/main/java/org/talend/sdk/component/proxy/service/ActionService.java
@@ -160,10 +160,16 @@ public class ActionService {
                     return referenceService
                             .findPropertiesById(configType, String.valueOf(it.getValue()), context)
                             .thenApply(form -> {
-                                final String prefix = it.getKey().replace(".$selfReference", ".");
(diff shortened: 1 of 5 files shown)


component-form/component-server-proxy-parent/component-server-proxy/src/main/java/org/talend/sdk/component/proxy/service/ActionService.java
```diff
@@ -157,14 +157,22 @@
...
+                    return referenceService
+                            .findPropertiesById(configType, String.valueOf(it.getValue()), context)
+                            .thenApply(form -> {
+                                final String prefix = it.getKey().replace(".$selfReference", "");
+                                synchronized (newProps) {
+                                    ofNullable(form.getProperties())
+                                            .ifPresent(props -> props
+                                                    .entrySet()
+                                                    .stream()
+                                                    .filter(k -> k.getKey().indexOf(".") > 0)
```
bchen-talend: is there any other way to filter $formId and find the root path properly?

rmannibucau: yep, idea is to find the form spec (simple property definition list) from the $formId of the nested config (form) and find the property where name==path

@akhabali is looking that since he is blocked by that as well

PR #337, opened 2018-11-06
ignore null values form array

akhabali:
### Requirements


rmannibucau:
Can need some love on js side since we should get payload without nulls normally // cc @jsomsanith

PR #342, opened 2018-11-19
feat(TFD-5795): Add a quick sample limit to PartitionMapper in Beam.

RyanSkraba:
### Requirements

### Why this PR is needed?

Provide quick sampling on PartitionMapper created through tacokit in the Beam transformation.

1. Limiting the number of sources/partitions to 1 when sampling is used, just sequentially asking the component for the next record.
2. "Countdown" for remaining records means that the limit is "per-source" (unlike Beam's BoundedSourceFromUnboundedSource, which reshuffles results).
3. Infinite mappers are turned into finite mappers when this stopping constraint is added.


RyanSkraba:
@rmannibucau suggests doing this in Streams instead of tacokit, and that the split might be necessary at least once on wrapped PartitionMapper instances (minimize splits by settting requested size to total size).

PR #346, opened 2018-12-11
feat(TCOMP-1229): support List<Enum>

bchen-talend:
### Requirements

### Why this PR is needed?
make multiSelectTag support chose from fixed list
to support List\<Enum\> for component configuration


akhabali:
How do we handle @Proposable trigger for this kind of list. We need also to handle the @Suggestable. for all list values or add a validation rules to prevent user that's it not supported for this type.
I think this is way tacokit api don't provide the List<Enum>.

can't you juste use the @Proposable to provide your Enum values as key,value.
here is the doc for it : https://talend.github.io/component-runtime/main/1.1.4/gallery.html#_list

rmannibucau:
Please ensure there are tests for these changes (and a manual ui test is done) + studio integration PR is ready to merge at the same time

rmannibucau:
no more needed, closing since it requires more work which will likely not be done now to not break the web ecosystem and in particular ComponentForm

PR #349, opened 2019-01-10
fix(TFD-6207): Relax beam core artifact identification by filename.

RyanSkraba:
### Requirements

### Why this PR is needed?

`beam-sdks-java-core` classes are identified and loaded from the parent classloader by looking at the code location, and matching against the filename of the jar.

Some of our services use the sbt-native-packager which copies artifacts to a `lib/` directory and *renames* them by prefixing the group id, which doesn't work with the current logic.

Relax how the beam core jar is identified by looking at the jar file name with and without the prefixed group id.

### DIFF
--- a/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
+++ b/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
@@ -885,7 +885,7 @@ public class ComponentManager implements AutoCloseable {
                         try {
                             return new File(decode(new URL(file.substring(0, separator)).getFile()))
                                     .getName()
-                                    .startsWith("beam-sdks-java-core-");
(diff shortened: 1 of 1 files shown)


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -885,7 +885,7 @@ protected boolean isContainerClass(final Filter filter, final String name) {
                         try {
                             return new File(decode(new URL(file.substring(0, separator)).getFile()))
                                     .getName()
-                                    .startsWith("beam-sdks-java-core-");
+                                    .contains("beam-sdks-java-core-");
```
rmannibucau: I assume it is due to the mutating packaging in your app which likely uses org.apache.beam.beam-sdks-java-core? If so can we use this as || startsWith instead of using contains which is more fragile?

PR #352, opened 2019-01-18
TCOMP-1265 : allow nested datastore and dataset in component configur…

akhabali:
…ation model

### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-1265

### DIFF
--- a/component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
+++ b/component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
@@ -821,7 +821,12 @@ public class ComponentValidator extends BaseTask {
     }
 
     private Stream<ParameterMeta> flatten(final Collection<ParameterMeta> options) {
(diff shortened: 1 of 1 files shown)


component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
```diff
@@ -821,7 +821,12 @@ private boolean isOutput(final Class<?> component) {
     }
 
     private Stream<ParameterMeta> flatten(final Collection<ParameterMeta> options) {
-        return options.stream().flatMap(it -> Stream.concat(Stream.of(it), it.getNestedParameters().stream()));
+        return options
+                .stream()
+                .flatMap(it -> Stream
+                        .concat(Stream.of(it),
+                                (it.getNestedParameters() == null || it.getNestedParameters().isEmpty()) ? empty()
```
rmannibucau: it will never be null so you can skip this test + please add a test for it

PR #359, opened 2019-02-25
chore : add an option to disable AllowedIPs check

amarouni:
### Requirements

### Why this PR is needed?
There's an inconsistent behavior when the proxy is deployed in a docker stack with a user defined netwrok (weave, ...). This PR allows the user to disable completely the AllowedIPs check. 

A new option with a default False value to disable the AllowedIPs check

### DIFF
--- a/component-server-parent/component-server-vault-proxy/src/main/java/org/talend/sdk/component/runtime/server/vault/proxy/endpoint/security/SecurityFilter.java
+++ b/component-server-parent/component-server-vault-proxy/src/main/java/org/talend/sdk/component/runtime/server/vault/proxy/endpoint/security/SecurityFilter.java
@@ -58,9 +58,9 @@ public class SecurityFilter implements Filter {
     private Boolean docker;
     
     @Inject
-    @Documentation("Disable IP/host white list")
-    @ConfigProperty(name = "talend.vault.cache.security.allowedIps.disable", defaultValue = "false")
-    private Boolean disableAllowedIps;
+    @Documentation("Enable Allowed IP list")
+    @ConfigProperty(name = "talend.vault.cache.security.allowedIps.enable", defaultValue = "true")
+    private Boolean enableAllowedIPsCheck;
 
     @Inject
     private DockerHostNameSanitizer dockerSanitizer;
@@ -90,11 +90,11 @@ public class SecurityFilter implements Filter {
 
     // cheap security
     private boolean isAllowed(final HttpServletRequest request) {
-        if(!disableAllowedIps) {
+        if(enableAllowedIPsCheck) {
+            return allowedIp.contains(request.getRemoteAddr()) || allowedIp.contains(sanitizeHost(request.getRemoteHost()));
+        } else {
             log.debug("Allowed IPs checking is disabled, request will be accepted");
             return true
-        } else {
-            return allowedIp.contains(request.getRemoteAddr()) || allowedIp.contains(sanitizeHost(request.getRemoteHost()));
         }
     }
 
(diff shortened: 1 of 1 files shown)


rmannibucau:
Has been fixed with hostbame sanitizer which enables to work with hosts in docker with or without weave when lookups are enabled.

RyanSkraba:
This would be useful in dev -- the sanitizers (docker and weave) depend on the version of docker-compose.  It doesn't work for docker-compose v1.23.0 or v1.23.1, for example, where the containers are named `<project>_<service>_<index>_<slug>` https://docs.docker.com/release-notes/docker-compose/#1230-2018-10-30

rmannibucau:
No more needed AFAIK
Security system will need to be revisited anyway at a later stage when we will change the deployment so closing for now until it happens

sgandon:
Please could you re-open this PR and for kubernetes it is impossible to know the IP or the host name of the callers due to the dynamic nature of K8s.
More over K8s provides such a security feature using NetworkPolicies.
So in ordre to make the component server we need to disable this check.

rmannibucau:
@sgandon security can be configured with Authorization header with a k8s secret. Network security was not intended to be used except in dev and for testing purposes so it matches your request already.

sgandon:
@rmannibucau thanks for your answer.
> Network security was not intended to be used except in dev and for testing purposes so it matches your request already.
If you are talking about K8s Network Policies then no it is not only for dev or testing and very much only for production workloads.
>security can be configured with Authorization header with a k8s secret
Can you elaborate this is not clear. I don't see the relation between the ip check and the secrets.

rmannibucau:
@sgandon idea is to use a real token based on K8s or equivalent (potentially vault which is integrated to the configuration) to secure the calls per header (as "any app") to avoid network whitelisting in env where it is not possible, here is a sample https://talend.github.io/component-runtime/main/1.1.20/vault-proxy.html#_docker_compose.

sgandon:
So what you suggest is to use this property `talend.vault.cache.security.tokens` to specify a token that shall be used by the caller as a header to authentify.
So this means means we need to change all the callers to add this token in the header  and share the token between all apps. 
Well this seems overly complicated when you can easilly rely on NetworkPolicies which is done precisely for this purpose.

rmannibucau:
@sgandon it is part of the setup of the proxy (this + the SSL setup when not colocalized with the server), it is already possible to disable that but it is not documented and requires some config in the deployment. Since there is already a trivial solution to secure the communication I decided to avoid to open an explicit security breach in the code which would be used outside secured networks for sure. I'm sure setting a secret is not that long and not worth this toggle for the mentionned reason but up to you.

sgandon:
The trouble is not creating a secret, it is adapting all apps calling the component server to add a new header with the token and provide the token to the apps too.
2 questions then, when you say 
>the SSL setup when not colocalized with the server

Do you mean between the proxy and the component-server ? does this mean that if in the same pod we could remove the ssl setup ? in the same pod this is a localhost communication so ssl between 2 containers in the same pod is also overkill.

When you say 
>it is already possible to disable that but it is not documented and requires some config in the deployment.

Can you elaborate and give the solution to do that please ?

rmannibucau:
> it is adapting all apps calling the component server to add a new header with the token

Technically there can be only one to adapt (REclient)...that said it is like that since v1 so it is a bug in callers they must fix, nothing to discuss there IMHO.

> does this mean that if in the same pod we could remove the ssl setup

No, just that it can be done automatically cause server is able to generate a cert and locally it can be shared through local filesystem (please review the config and sample linked before, all is done there).

> Can you elaborate and give the solution to do that please ?

Not needed ;) but it server is designed to be extended (and it is) and there are hooks to drop/replace beans so doable  in ~5LOC but once again it is *not* a good solution for such deliverable and since an easy solution is built-in no real way to break it IMHO.

sgandon:
I completly disagree but I am tired to argue.

PR #360, opened 2019-03-14
[TCOMP-1341] Add DeployInStudioTask in Gradle Plugin

ZoltanTakacs:
## Why this PR is needed?
- Currently the Talend Gradle Component plugin lacks of the possibility to deploy the module components to the Studio.

## What does this PR adds (design/code thoughts)?
- The task was already existed in the plugin, but was not registered. I just added this task so users can deploy from CLI.

### DIFF
--- a/gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
+++ b/gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
@@ -82,9 +82,9 @@ public class TaCoKitPlugin implements Plugin<Project> {
         project.task(new HashMap<String, Object>() {
 
             {
-                put("type", DependenciesTask.class);
-                put("group", group);
-                put("description",
+                put(Task.TASK_TYPE, DependenciesTask.class);
+                put(Task.TASK_GROUP, group);
+                put(Task.TASK_DESCRIPTION,
                         "Creates the Talend Component Kit dependencies file used by the runtime to build the component classloader");
             }
         }, "talendComponentKitDependencies");
@@ -100,9 +100,10 @@ public class TaCoKitPlugin implements Plugin<Project> {
         project.task(new HashMap<String, Object>() {
 
             {
-                put("type", ValidateTask.class);
-                put("group", group);
-                put("description", "Validates that the module components are respecting the component standards.");
+                put(Task.TASK_TYPE, ValidateTask.class);
+                put(Task.TASK_GROUP, group);
+                put(Task.TASK_DESCRIPTION,
+                        "Validates that the module components are respecting the component standards.");
             }
         }, "talendComponentKitValidation");
         project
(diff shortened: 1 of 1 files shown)


ZoltanTakacs:
Hi @rmannibucau !
I'm not sure if this is the right process for contributing into this project, but wanted to send you a quick fix in the gradle plugin. Please let me know if you'd like me to change something.
- If you could add this fix in the next release of the plugin that would be awesome!

Best,
Zoltan

gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
```diff
@@ -149,5 +150,16 @@ public void apply(final Project project) {
                 put("description", "Creates a Component ARchive (.car) based on current project.");
             }
         }, "talendComponentKitComponentArchive");
+
+        // deploy in studio
+        project.task(new HashMap<String, Object>() {
+
+            {
+                put(Task.TASK_TYPE, DeployInStudioTask.class);
```
rmannibucau: On the style can you please use direct string as others or update others with these constants please?

gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
```diff
@@ -149,5 +150,16 @@ public void apply(final Project project) {
...
+                put(Task.TASK_GROUP, "Talend Component Kit deployment");
+                put(Task.TASK_DESCRIPTION, "Deploys the module components to the Studio.");
+                put(Task.TASK_DEPENDS_ON, "jar");
+            }
+        }, "deployInStudio");
```
rmannibucau: Can you add the same prefix than other tasks please for consistency and get a better output when using tasks task from the cli?

rmannibucau:
Hi @ZoltanTakacs

This is the perfect process!

I just put two style comments you should be able to fix quickly but sure we will get it in 1.1.8 release.

gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
ZoltanTakacs: Sure thing, updated all the occurrences to use constants. 👍

gradle-talend-component/src/main/java/org/talend/sdk/component/gradle/TaCoKitPlugin.java
ZoltanTakacs: Update this one as well to use the same prefix and group. 👍

ZoltanTakacs:
Hi @rmannibucau 

I've updated the PR with the suggested changes. :) If there is anything else, just let me know!

Thanks for reviewing it!

rmannibucau:
:+1: awesome, thanks a lot, it is in!

ZoltanTakacs:
Thanks very much @rmannibucau! 🎉🎆🙂

PR #370, opened 2019-05-02
fix(TCOMP-1411): Relax matching for any jar with beam.sdks.java.core.

RyanSkraba:
### Requirements

### Why this PR is needed?

Quick fix example for TCOMP-1411 -- better solution would be welcome.

### DIFF
--- a/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
+++ b/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
@@ -965,8 +965,7 @@ public class ComponentManager implements AutoCloseable {
                         try {
                             final String strippedJar =
                                     new File(decode(new URL(file.substring(0, separator)).getFile())).getName();
-                            return strippedJar.startsWith("beam-sdks-java-core-")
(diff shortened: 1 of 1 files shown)


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -965,8 +965,7 @@ protected boolean isContainerClass(final Filter filter, final String name) {
                         try {
                             final String strippedJar =
                                     new File(decode(new URL(file.substring(0, separator)).getFile())).getName();
-                            return strippedJar.startsWith("beam-sdks-java-core-")
-                                    || strippedJar.startsWith("org.apache.beam.beam-sdks-java-core-");
+                            return strippedJar.matches(".*beam.sdks.java.core.*");
```
rmannibucau: Hi ryan

2 notes on that

1. Please no regex, it us too slow at startup
2. Why is it needed? (Assuming pd uses a configurer)

rmannibucau:
@RyanSkraba closing since TCOMP-1411 should address that. Let me know if still needed. Thanks.

PR #372, opened 2019-05-13
fix(TUP-20649), [TCOMP-1423]:Fail to install car file when set maven.repostiory=global.

zwxue:


zwxue:
Hello, this is for light commandline feature, it uses global maven repository by default, and it failed while trying to install car patch, so need to apply this change on Tacokit side.

rmannibucau:
Applied, thanks!

PR #374, opened 2019-05-16
Change uses of Beam guava from repackaged to vendored

iemejia:
Repackaging is being removed incrementally in favor of the vendored dependency.
Found this issue when trying to run with the new 2.14.0-SNAPSHOT

### DIFF
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/coder/registry/AvroCoderCache.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/coder/registry/AvroCoderCache.java
@@ -21,9 +21,9 @@ import java.util.concurrent.ExecutionException;
 
 import org.apache.avro.Schema;
 import org.apache.avro.generic.IndexedRecord;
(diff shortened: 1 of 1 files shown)


rmannibucau:
Hi @iemejia , do you think it is possible to drop guava and not depend on beam anymore instead of using a not direct dependency which introduces a new link we likely want to avoid due to the instability there?

iemejia:
Sadly no, or at least not in a reasonable time. Getting guava out of Beam implies a HUGE effort that it is hard to justify, because we won't get new features from it, apart of course of cleaning the dependencies.

rmannibucau:
@iemejia was not speaking of beam here but more the PR itself

iemejia:
Well I don't knowm I send back the question to you, why it was chosen to use Guava in that code in the first place?

rmannibucau:
@iemejia cause it was there in direct dependencies, thread safe OOTB and trivial to use. Will drop it as part of TCOMP-1435.

iemejia:
Closing since it is already fixed.

PR #387, opened 2019-08-21
fix(TCOMP-1516): Use the ContainerFinder impl in component-server.

RyanSkraba:
### Requirements


rmannibucau:
Hi Ryan,
This will likely break other modules.
Did you investigate https://jira.talendforge.org/browse/TFD-8519?focusedCommentId=694523&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-694523 ?

Romain

rmannibucau:
Hi @RyanSkraba ,
your last commit does not help since it just makes tests passing but breaks all consumers (users to make it simple).
Can you try to implement the way I described in the jira which was the intended solution for the ticket? ContainerFinder is not intended to be used like that at all and it is risky since it will break other usages to let it be (see jira comment for details).

rmannibucau:
will be handled in pipeline designer integration based on framework SPI so closing this one in favor of the other work being done

thanks for sharing anyway, it is always better than words ;)

amarouni:
@RyanSkraba @undx 
This PR was rejected and closed last year for a number of reasons, so I suggest that we reopen it and discuss the solution and other possible alternatives.

RyanSkraba:
I took the liberty of reopening in order to rebase and rebuild.  We've been bumping the embarrassing workaround described in TFD-8518 for every version since!  This PR is a fairly simple solution worth revisiting.

PR #391, opened 2019-09-23
feat(TFD-8321): Fix invalid classloader.

pbailly:
### Requirements

### Why this PR is needed?
The current classloader duplicated avro-mapred jars. They need to be excluded.
There is also an invalid include on the package `org.apache.beam.repackaged` that only include `org.apache.beam.repackaged.core` but not think like `org.apache.beam.repackaged.beam_sdks_java_core`.

Fix the groovy file.

### DIFF
--- a/component-runtime-beam/src/main/build/ClassesIndices.groovy
+++ b/component-runtime-beam/src/main/build/ClassesIndices.groovy
@@ -92,6 +92,10 @@ def excludedPackages = [
     'org.mortbay.',
     'org.kohsuke.',
     // todo: check
+    'org.apache.avro.hadoop.',
+    'org.apache.avro.ipc.',
+    'org.apache.avro.mapred.',
+    'org.apache.avro.mapreduce.',
     'org.apache.zookeeper.',
     'org.apache.hadoop.',
     'org.apache.directory.',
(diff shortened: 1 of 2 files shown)


component-runtime-beam/src/main/build/ClassesIndices.groovy
```diff
@@ -92,13 +92,18 @@ def excludedPackages = [
     'org.mortbay.',
     'org.kohsuke.',
     // todo: check
+    'org.apache.avro.hadoop.',
```
rmannibucau: This is part of spark infra, we dont want to nlt filter them

component-runtime-beam/src/main/build/ClassesIndices.groovy
```diff
@@ -92,13 +92,18 @@ def excludedPackages = [
...
+    'org.apache.avro.mapred.',
+    'org.apache.avro.mapreduce.',
     'org.apache.zookeeper.',
     'org.apache.hadoop.',
     'org.apache.directory.',
     'com.codahale.metrics.'
 ]
 
 def simplifiedPackages = [
+    'org.apache.beam.repackaged', // we want everything inside repackaged
```
rmannibucau: Likely the wrong fix, guess https://github.com/pbailly/component-runtime/blob/ee4d7479fa79341d457fb11b22ad8694b99b8781/component-runtime-beam/src/main/build/ClassesIndices.groovy#L155 should let classes pass with this end dot so likely one of the previous substring messes up thing
Will have a look later this week since all that code will need to be adjusted for 1.1.13 anyway

rmannibucau:
fixed with https://jira.talendforge.org/browse/TCOMP-1202

PR #396, opened 2019-12-20
Add configuration to allow directly a subnet on vault-proxy

lcelton-talend:
### Requirements

### Why this PR is needed?
In the context of K8s (cluster dev), it's not easy to configure the security to allow the communication between the client app and vault-proxy. We don't want to add a header for any request to the tacokit-vault-proxy. 

The solution is to add a new configuration to the vault-proxy to allow a subnet and keep the security without adding any code in the client app.

### DIFF
--- a/component-server-parent/component-server-vault-proxy/pom.xml
+++ b/component-server-parent/component-server-vault-proxy/pom.xml
@@ -42,6 +42,11 @@
       <version>${commons-cli.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


rmannibucau:
We want to fully control the connected machines so network enablement is either explicit (machine) or forbidden and the auth use a token so I am -1 for that PR but happy to drop network security which was only there for convenience/local dev and not really prod friendly.

rmannibucau:
Closing since the other - right - solution is being worked on.

PR #421, opened 2020-02-14
TCOMP-1643 - Add automatic scheduling eviction system on LocalCache

clesaec:
### Requirements

### Why this PR is needed?

LocalCache with time-out eviction.

LocalCache interface and LocalCacheService modifications.


component-api/src/main/java/org/talend/sdk/component/api/service/cache/LocalCache.java
```diff
@@ -15,51 +15,69 @@
...
 /**
  * Framework service injectable in components or services methods
  * to handle local caching.
  * Useful for actions when deployed in a long running instance
  * when actions are costly.
+ * 
+ * @param <T>  the type of cached data.
  */
-public interface LocalCache {
+public interface LocalCache<T> extends Serializable {
```
rmannibucau: This cache is not typed so better to not type it since it can end up with a classcastexception by design.

component-api/src/main/java/org/talend/sdk/component/api/service/cache/LocalCache.java
```diff
@@ -15,51 +15,69 @@
...
     /**
-     * Remove a cached entry.
-     *
-     * @param key key to evict.
+     * Use to enrich object with meta-data (help tio choice if to be removed)
+     * @param <T> the type of data to cache.
      */
-    void evict(String key);
+    interface Element<T> {
+        T getValue();
```
rmannibucau: Likely replace by a T getValue(Class<T> expectedType) instead

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
     @Override
-    public <T> void evictIfValue(final String key, final T expected) {
-        cache.remove(internalKey(key), new Element(expected, 0));
+    public void evictIfValue(final String key, final T expected) {
+        final String realKey = internalKey(key);
+
+        // use compute to be able to call release.
+        cache.compute(realKey,
+                (String oldKey , ElementImpl<T> oldElement) -> {
+                    if (oldElement != null && Objects.equals(oldElement.getValue(), expected)) {
```
rmannibucau: I suspect it can be neat to benefit the fact you have oldElement to check also the timestamp if not null to do a normal eviction

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+
+
     @Override
-    public <T> T computeIfAbsent(final String key, final long timeoutMs, final Supplier<T> value) {
+    public T computeIfAbsent(String key,
+            final Predicate<Element<T>> toRemove,
+            final Supplier<T> value) {
+
+        final ElementImpl<T> element = this.addToMap(key,
+                () -> new ElementImpl<>(value, toRemove, -1, null));
```
rmannibucau: shouldn't it get an optional scheduled task too using the configuration timeout?

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
 
     @Override
-    public <T> void evictIfValue(final String key, final T expected) {
-        cache.remove(internalKey(key), new Element(expected, 0));
+    public void evictIfValue(final String key, final T expected) {
+        final String realKey = internalKey(key);
+
+        // use compute to be able to call release.
+        cache.compute(realKey,
+                (String oldKey , ElementImpl<T> oldElement) -> {
```
rmannibucau: types an be removed from the lambda definition lighter def)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
-                        (k, e) -> e == null || e.isExpired()
-                                ? new Element(value.get(), System.currentTimeMillis() + timeoutMs)
-                                : e);
-        if (element == null) {
-            return null;
-        }
-        return (T) element.value;
+        return cache.compute(
+                internalKey,
+                (String k, ElementImpl<T> old) ->    //
```
rmannibucau: types an be removed from the lambda definition lighter def)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+    public boolean isEmpty() {
+        return this.cache.isEmpty();
+    }
+
+    private ScheduledExecutorService getThreadService() {
+        if (this.threadService == null) {
+            synchronized (this) {
+                if (this.threadService == null) {
+                    this.threadService = Executors.newSingleThreadScheduledExecutor((Runnable r) -> {
+                        final Thread thread = new Thread(r, this.getClass().getName() + "-eviction-" + hashCode());
```
rmannibucau: this getClass() will not be the one you want (likely the lambda one), maybe reference the LocalCacheService.class.getName() instead, it should also get the plugin to differenciate instances per family

side note: best would be to have a single global default evictor when config is the same to avoid to create an uncontrolled number of thread (x#families which can be a lot in the server) IMO

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+
+        /** give time object can be release (infinity if < 0) */
         private final long endOfValidity;
 
-        private boolean isExpired() {
-            return System.currentTimeMillis() > endOfValidity;
+        /** scheduled task to remove object if nedeed (to cancel if removed before) */
+        private final ScheduledFuture<?> removedTask;
+
+        public ElementImpl(Supplier<T> value,
```
rmannibucau: finals ;)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static java.util.stream.Collectors.toList;
 
+
+/**
+ * Implementation of LocalCache with in memory conccurent map.
+ * @param <T>   the type of cached data.
+ */
+@RequiredArgsConstructor
+public class LocalCacheService<T> implements LocalCache<T>, Serializable {
```
rmannibucau: misses a close callback which stops the executor if not null (so this leaks threads)

Think a @PreDestroy callback should work but ensure it with a test (checking threads before creating the local cache service, while it is active (+1 thread) and after the destruction (back to the same as before the creation)

Finally it will require to handle the state of the service (closed or not) to prevent the pool to be recreated and eviction to be submitted if the service is already closed.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -15,46 +15,167 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.talend.sdk.component.runtime.manager.test.Serializer.roundTrip;
+
 class LocalCacheServiceTest {
 
+
+    private boolean active;
+
+    private int  intervale;
```
rmannibucau: -e ;)

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -15,46 +15,167 @@
...
+                return String.valueOf(intervale);
+            }
+            if ("test.talend.component.manager.services.cache.eviction.active".equals(key)) {
+                return String.valueOf(active);
+            }
+            throw new IllegalArgumentException("unknown key '" + key + "'");
+        }
+
+        @Override
+        public Set<String> keys() {
```
rmannibucau: not needed, note that you can likely reuse org.talend.sdk.component.runtime.manager.service.configuration.PropertiesConfiguration and handle the configuration in a Properties instead instead

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -15,46 +15,167 @@
...
+    void init() {
+        this.active = false;
+        this.intervale = 0;
+        final Map<Class<?>, Object> services = new HashMap<>(2);
+       // services.put(LocalCache.class, new LocalCacheService("LocalCacheServiceTest"));
+        services.put(LocalConfiguration.class,
+                new LocalConfigurationService(Collections.singletonList(cacheConfig), "test"));
+
+
+        final PropertyEditorRegistry propertyEditorRegistry = new PropertyEditorRegistry();
```
rmannibucau: not sure this helps the test, at the end it just gets back the instance you created just before

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -15,46 +15,167 @@
...
+    }
+
+    @Test
+    void localTimeout() throws InterruptedException {
+        final String v1 = cache.computeIfAbsent("key1", 200L, () -> "value1");
+        final String v2 = cache.computeIfAbsent("key1", 200L,  () -> "value2");
+
+        final String valueKey2 = cache.computeIfAbsent("key2", -1L, () -> "valueKey2");
+
+        Thread.sleep(240L);
```
rmannibucau: I'd increase a bit the latency, windows is bad on that so likely +-100ms is not luxury (you can increase the timeout (200) if you judge the % of the error is too close to the value)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+        final ElementImpl<T> element = this.addToMap(key,
+                () -> new ElementImpl<>(value, toRemove, -1, null));
+
+        return element.value;
+    }
+
+    @Override
+    public T computeIfAbsent(String key, long timeoutMs, Supplier<T> value) {
+
+        final long endOfValidity = timeoutMs > 0 ? System.currentTimeMillis() + timeoutMs : -1;
```
rmannibucau: we should likely move to Clock, this would enable to test it moreeasily
In the default service provider it can be the system utc default one but in test you will pass the "test"/settable one.
Will avoid the "sleeps".

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,236 @@
...
+        final List<String> removableElements = this.cache
+                .entrySet()
+                .stream()
+                .filter(e -> e.getValue().mustBeRemoved())
+                .map(Entry::getKey)
+                .collect(toList());// materialize before actually removing it
+        removableElements.forEach(this.cache::remove);
+    }
+
+    public boolean isEmpty() {
```
rmannibucau: not really needed and will impact proxying for nothing IMHO, this is a test thing, just do reflection in the test?

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: ??? in this function, we force the eviction whatever timestamp value.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
clesaec: keeping LocalConfiguration like this allow me to have dynamic value (active, interval) for unit test

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: Number of thread is defined by threadService class :
this.threadService = Executors.newScheduledThreadPool(4, (Runnable r) -> ..) 
will never go over 4 thread, only the Queue that will have numbers of Runnable.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
clesaec: I just followed java API to inject value to services. If you suggest better, i can do it.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
rmannibucau: same using the properties, you just do "setProperty" instead of "=" and it reduces the boilerplate significatively IMHO

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
rmannibucau: but you don't test the injection, you test the instance you injected so, until I misunderstood the test, it does not do anything for this particular test :thinking:

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
rmannibucau: if it is equals, when it is not equals (so the cached value is not the expected one for that key) the proposal was to check the timestamp to do the eviction without waiting for the background thread (which can be inactive in this env)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
rmannibucau: do you think? this number is per family using local cache (so almost all) so it will quickly grow if not factorised with a default global pool (lazily created too) when no specific configuration is set.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: ok, get it !

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: So, PreDestroy should also be for this global pool (shutdownNow & awaitTermination).
How to have one global pool for several LocalCache (@ Inject => but not lazy, @ Service, Add to constructor or static instance ?)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: All Clock subclasses are immutable, so i can't modify time.
With a mutable Clock class, i can easily play with endOfValidity field.
But how with ScheduledExecutorService.schedule ?

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: removed

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
clesaec: removed

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
clesaec: done, even if it's useless on method parameter.

component-api/src/main/java/org/talend/sdk/component/api/service/cache/LocalCache.java
```diff
@@ -26,40 +27,93 @@
 public interface LocalCache {
 
     /**
-     * Remove a cached entry.
-     *
-     * @param key key to evict.
+     * Use to enrich object with meta-data (help tio choice if to be removed)
```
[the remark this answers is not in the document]
clesaec: '(help to choice if attached object has to be removed from cache)'
Better ?

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/DefaultServiceProvider.java
```diff
@@ -251,4 +260,19 @@ private Properties aggregateConfigurations(final Stream<InputStream> localConfig
...
+
+    /**
+     * Build executor service
+     * used by
+     * - Local cache for eviction.
+     * 
+     * @return scheduled executor service.
+     */
+    private ScheduledExecutorService buildExecutorService() {
+        return Executors.newScheduledThreadPool(4, (Runnable r) -> {
```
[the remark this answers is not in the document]
clesaec: How ? it's a 6 lines method. 
So very simple comparing to methods with more than 200 lines ...

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,258 @@
...
+        this.cache.clear();
+    }
+
+    private long calcEndOfValidity(final long timeoutMs) {
+        return timeoutMs > 0 ? this.timer.get() + timeoutMs : -1;
+    }
+
+    private long getDefaultTimeout() {
+        final CacheConfiguration config = getConfig();
+        return config != null && config.isActive() ? config.getDefaultEvictionInterval() : -1;
```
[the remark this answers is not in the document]
clesaec: Yes, that's the meaning.
Other programming strategy should to build a class for TimeOut with long and a boolean (or enum), cleaner but heavier. So, i chose light one.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -15,45 +15,214 @@
...
+        assertEquals("bar", v2);
+        try {
+            latch.await();
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+        final long end = System.currentTimeMillis();
+
+        final long duration = end - start;
+        assertEquals(200, duration, 100);
```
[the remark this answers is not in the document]
clesaec: changed to 200 "assertEquals(200, duration, 200);", better ?

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,258 @@
...
+    }
+
+    /**
+     * Cache configuration.
+     */
+    @Data
+    public static class CacheConfiguration implements Serializable {
+
+        @Option
+        private long defaultEvictionInterval;
```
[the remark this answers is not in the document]
clesaec: changed to defaultEvictionTimeout

component-api/src/main/java/org/talend/sdk/component/api/service/cache/LocalCache.java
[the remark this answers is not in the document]
undx: crystal clear ;-)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/DefaultServiceProvider.java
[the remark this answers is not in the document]
undx: no was just to say it could written `r ->`. A matter of style.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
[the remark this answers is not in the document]
undx: :+1:

clesaec:
> I suggest to create another test with a test component having a @cached method and
> a Localcache @service in it. So, you could make a real test as planned to be used.

There is already this [interceptor test](https://github.com/Talend/component-runtime/blob/master/component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/interceptor/InterceptorTest.java#L56) that does it. 
I can enrich the test to test automated eviction (but already test in first)
For Workday, i didn't plan to use '@Cached' annotation until it's enrich (time out is not a constant but come from workday metadata service response)

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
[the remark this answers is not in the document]
undx: Think that it should pass, on my workstation the delta was around 140... Just making sure that tests pass all the time. ;-)

undx:
> > I suggest to create another test with a test component having a @cached method and
> > a Localcache @service in it. So, you could make a real test as planned to be used.
> 
> There is already this [interceptor test](https://github.com/Talend/component-runtime/blob/master/component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/interceptor/InterceptorTest.java#L56) that does it.
> I can enrich the test to test automated eviction (but already test in first)
> For Workday, i didn't plan to use '@cached' annotation until it's enrich (time out is not a constant but come from workday metadata service response)

Yes, I think that you can enrich the existing test with new use-cases and features like eviction with  a predicate for instance.

undx:
> * in images/component-server-image the system property must be set to have a default eviction timeout (likely 30s)
> * Think it misses a "maxDeletionPerEvictionRun" parameter too
> * in prod the size() should be used to do eviction as a safe guard, at read time (if > defaultMaxSize=1000 then trigger an eviction - still with maxDeletionPerEviction param)
Just a remimder on Romain's changes and recomandations request.

clesaec:
About 'enrich interceptor test` : for the moment, we keep @Cached annotation as it, only with hard coded timeout. We may do evolution on it on a future release.

Romain reco :
images/component-server-image : Help me ... Where do i put default eviction timeout.

in prod the size() should be used to do eviction as a safe guard, at read time (if > defaultMaxSize=1000 then trigger an eviction ...) : So i add a defaultMaxSize to CacheConfig. If, after a clean, size is still over defaultMaxSize ... do i remove element randomly until i reach correct size ?

undx:
> Romain reco :
> images/component-server-image : Help me ... Where do i put default eviction timeout.

https://github.com/Talend/component-runtime/blob/master/images/component-server-image/pom.xml#L219

undx:
> About 'enrich interceptor test` : for the moment, we keep @cached annotation as it, only with hard coded timeout. We may do evolution on it on a future release.

Make maybe another test based on this one but testing your advanced features...

undx:
> in prod the size() should be used to do eviction as a safe guard, at read time (if > defaultMaxSize=1000 then trigger an eviction ...) : So i add a defaultMaxSize to CacheConfig. If, after a clean, size is still over defaultMaxSize ... do i remove element randomly until i reach correct size ?

I'll say that we will plan this feature for later. Meanwhile, we'll discuss on it.
For now, the localcache is not very used so we can delay it a _bit_.

clesaec:
All requested features are done :-)

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
```diff
@@ -64,17 +242,78 @@ void serialize() throws IOException, ClassNotFoundException {
...
+        Assertions.assertTrue(isCacheEmpty(), "not empty after last clean");
+    }
+
+    @Test
+    public void evictionAuto() {
+        this.defaultMaxSize = 10;
+        for (int i = 0; i < 20; i++) {
+            cache.computeIfAbsent(String.class, "k" + i, (Element e) -> false, -1L, () -> "val");
+        }
+    }
```
undx: No assertion in this test.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,285 @@
...
-    public static class Element {
+    public void clean() {
+        Stream<Entry<String, ElementImpl>> elements =  //
+                this.cache.entrySet() //
+                        .stream()     //
+                        .filter(e -> e.getValue().mustBeRemoved());
+
+        final int maxEviction = this.getConfigValue(CacheConfiguration::getMaxDeletionPerEvictionRun, -1);
+        if (maxEviction > 0) {
+            elements = elements.limit(maxEviction);
```
undx: If we select a subset of removable items maybe we need to define a sort strategy... Oldest first, nearest timeout or something like that.

clesaec: sort strategy will necessary impact performances O(N.log(N)) vs Constant. As it's 'only' a cache, i prefer keep good performances.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
```diff
@@ -17,71 +17,285 @@
...
+        if (maxSize > 0 && this.cache.size() >= maxSize) {
+            this.clean(); // clean before add one element.
+            if (this.cache.size() >= maxSize) {
+                synchronized (this.cache) {
+                    while (this.cache.size() >= maxSize) {
+                        final String keyToRemove = this.cache.keySet().iterator().next();
+                        this.cache.remove(keyToRemove);
+                    }
+                }
+            }
```
undx: If maxSize is reached, you first do a clean() which removes all keys that can be removed but which can be limited by the MaxDeletionPerEvictionRun (which has sense) ... but after (if still over the maxsize allowed) you loop unconditionnaly on all keys and remove them one by one ?
Not sure about this process... Can we think a task that will run and do this ? 
(Just a proposal to be studied)
Below, you create a task for the eviction for each key which has a timeout defined...
Why not having only one task which will do the maintenance? 
ie : eviction of expired keys and check the cache size

clesaec: - The clean is inside the 'adding' function. So, if it oversize maxSize; it should be small oversizing. So, if on line 113, it's still oversize maxSize; then, there is a strong probability that it's because items have either a valid date either a method tell we can keep object. So, the one by one eviction until we reach MaxSize seems to me a good strategy.

think a task : You mean a task that run at scheduled time (risk is to remove last element) or put the loop in separated thread ? Second option will allow to optimize insertions if creation of new task is quicker than clean method.

a task for each key allow to remove old object without scanning all cache map.

undx: hum, not sure about the strategy but as commented before, we now can stick to this one and advice a bit later.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/LocalCacheService.java
undx: Same comment as for the clean and maxsize thread.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/LocalCacheServiceTest.java
clesaec: added

PR #425, opened 2020-04-01
TCOMP-1692 - Update CXF to 3.3.6

coheigea:
We should update Apache CXF to 3.3.6 due to CVE-2020-1954:

http://cxf.apache.org/security-advisories.data/CVE-2020-1954.txt.asc?version=1&modificationDate=1585730169000&api=v2


coheigea:
That was fast, thanks @undx

PR #432, opened 2020-04-29
feat(TCOMP-1709) : Record should return Object notprimitive types

ypiel-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-1709
we couldn't return null with getDouble() for example.

Record getters return Objects that can be null


rmannibucau:
Hi Yves,
This is a breaking change + nullable values must use optional accessors so looks like it introduces issues in component and breaks API contract ;)

PR #443, opened 2020-07-17
wi(TFD-10924):Heterogeneous schema - step 1 Misaligned/missing columns like Workday connector case

xtao-talend:
### Requirements


undx:
Closing as jira issue is closed.

PR #445, opened 2020-09-18
TCOMP-1771 : record builder optim

clesaec:
### Requirements
JIRA => https://jira.talendforge.org/browse/TCOMP-1771

### Why this PR is needed?
basic optimization 
no design added, code optim & tested.


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -115,25 +115,37 @@ public BuilderImpl(final Schema providedSchema) {
...
                 throw new IllegalArgumentException(
                         "No entry '" + name + "' expected in provided schema: " + entryIndex.keySet());
             }
-            if (entry.getType() != type) {
-                throw new IllegalArgumentException(
-                        "Entry '" + name + "' expected to be a " + entry.getType() + ", got a " + type);
+            return entry;
+        }
+
+        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type, boolean nullable) {
```
undx, line 131: `nullable` should be f`final`.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -115,25 +115,37 @@ public BuilderImpl(final Schema providedSchema) {
...
         }
 
-        private void validateTypeAgainstProvidedSchema(final String name, final Schema.Type type, final Object value) {
-            if (providedSchema == null) {
-                return;
-            }
-            if (entryIndex == null) {
-                entryIndex = providedSchema.getEntries().stream().collect(toMap(Schema.Entry::getName, identity()));
+        private Schema.Entry findMandatoryEntry(final String name) {
+            assert this.providedSchema != null;
```
undx, line 119: why using `assert` here?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -115,25 +115,37 @@ public BuilderImpl(final Schema providedSchema) {
             this.providedSchema = providedSchema;
         }
 
-        private void validateTypeAgainstProvidedSchema(final String name, final Schema.Type type, final Object value) {
-            if (providedSchema == null) {
-                return;
-            }
-            if (entryIndex == null) {
-                entryIndex = providedSchema.getEntries().stream().collect(toMap(Schema.Entry::getName, identity()));
+        private Schema.Entry findMandatoryEntry(final String name) {
```
undx, line 118: function name: remove mandatory or rename it ...

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -115,25 +115,37 @@ public BuilderImpl(final Schema providedSchema) {
...
+        }
+
+        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type, boolean nullable) {
+            if (providedSchema == null) {
+                return new SchemaImpl.EntryImpl.BuilderImpl().withName(name).withType(type).withNullable(nullable).build();
+            }
+            return this.findMandatoryEntry(name);
+        }
+
+        private Schema.Entry validateTypeAgainstProvidedSchema(final String name, final Schema.Type type, final Object value) {
```
undx, line 138: Here, 
- the function name do not match the goal of the function and provided type is useless.
- the function has a side effect by creating a new entry and it doesn't match initial goal.

undx (commented):
see my comments

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: this function is private and must be called only when providedSchema exists. assert is normally not tested on running code, just in unit test (until we used -ea option)

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: renamed findExistingEntry, better ? why mandatory was not good ?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: done

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: - type checking added (don't know if it's usefull, assertType do also this control)
- this method never create entry.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
undx: sorry a matter of taste... ;-)

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
undx: still think that its place in in tests not here...

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
undx: you're right on second point ;-)

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: assert removed

undx (approved):
LGTM

PR #446, opened 2020-09-24
feat(TCOMP-1761): Support of complete schema definition in Studio

undx:
**Support of complete schema definition in Studio**
https://jira.talendforge.org/browse/TCOMP-1761

**Related PR**
- https://github.com/Talend/component-runtime/pull/446
- https://github.com/Talend/tdi-studio-se/pull/5270


component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
```diff
@@ -801,6 +801,12 @@ private boolean isListString(final Field f) {
                 && pt.getActualTypeArguments().length == 1 && pt.getActualTypeArguments()[0] == String.class;
     }
 
+    private boolean isListObject(final Field f) {
```
clesaec, line 804: you could directly test for collection :  

```java
   public boolean isCollectionOf(final Field f, Class<?> typeArg) {
        final ParameterizedType pt = ParameterizedType.class.cast(f.getGenericType());
        Type classType = pt.getRawType();
        if (pt.getActualTypeArguments().length != 1) {
            return false;
        }
        Type innerType = pt.getActualTypeArguments()[0];
        return classType instanceof Class<?> 
           && Collection.class.isAssignableFrom((Class<?>) classType) 
           && innerType instanceof Class<?> 
           && ((Class<?>) innerType).isAssignableFrom(typeArg);
    }
```

With this; a declared field `ArrayList<String>` function return true for  `isCollectionOf(field, String.class);`  but false for a field  `List<Object>`
with field `List<Object>`, function return true for `isCollectionOf(field, String.class);` (you can add a String to a List<Object> collection).

undx: Good point Chris.

ypiel-talend (approved):
LGTM

component-tools/src/main/java/org/talend/sdk/component/tools/validator/ModelValidator.java
```diff
@@ -102,15 +100,10 @@ private int countParameters(final Parameter[] params) {
...
-        return (Map.class == pt.getRawType()) && pt.getActualTypeArguments().length == 2
-                && pt.getActualTypeArguments()[0] == String.class && pt.getActualTypeArguments()[1] == String.class;
-    }
-
-    private boolean isListString(final Field f) {
+    private boolean isListObject(final Field f) {
         final ParameterizedType pt = ParameterizedType.class.cast(f.getGenericType());
         return ((List.class == pt.getRawType()) || (Collection.class == pt.getRawType()))
-                && pt.getActualTypeArguments().length == 1 && pt.getActualTypeArguments()[0] == String.class;
+                && (pt.getActualTypeArguments().length == 1) && (pt.getActualTypeArguments()[0] instanceof Object);
```
clesaec, line 106: A Collection (or List) will always have arguments type that are object instance (List is transform to List<Object>)
Why not accept "pt.getRawType() as a subtype of Iterable" (Iterable.isAssignableFrom(pt.getRawType())

instanceof with a type ? more Object.class.isAssignableFrom(xxx) => both case will always true

clesaec (commented):
comment here

PR #447, opened 2020-10-06
feat(TCOMP-1768): context variables migration

undx:
https://jira.talendforge.org/browse/TCOMP-1768


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+
+        listeners.forEach(e -> e.onAddKey(configuration, key, value));
+    }
+
+    /**
+     * @param oldKey configuration key existing
+     * @param newKey configuration key to rename
+     */
+    public final void renameKey(final String oldKey, final String newKey) {
+        configuration.put(newKey, configuration.get(oldKey));
```
ypiel-talend, line 91: Check if oldKey already exist, if not, throw an exception

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+        log.debug("[unRegisterListener] unregistering {}.", listener.getClass().getName());
+        listeners.remove(listener);
+    }
+
+    /**
+     * @param key
+     * @param value
+     */
+    public final void addKey(final String key, final String value) {
+        configuration.put(key, value);
```
ypiel-talend, line 81: Check if Key already exist, if yes, throw an exception

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+        configuration.remove(oldKey);
+
+        listeners.forEach(e -> e.onRenameKey(configuration, oldKey, newKey));
+    }
+
+    /**
+     * @param key configuration key
+     */
+    public final void removeKey(final String key) {
+        configuration.remove(key);
```
ypiel-talend, line 101: Check if key exists, if not throw an exception

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+
+        listeners.forEach(e -> e.onRemoveKey(configuration, key));
+    }
+
+    /**
+     * @param key configuration key
+     * @param newValue new value to set in migration
+     */
+    public final void changeValue(final String key, final String newValue) {
+        final String oldValue = configuration.get(key);
```
ypiel-talend, line 111: could we have a Function instead of String for newValue ? but we let a string for the listener. so we could have something lilke:
public final void changeValue(final String key, final Function<String, String> func) {
	String new value = func.apply(config.get(key))
	configuration.put(key, newValue);
	listeners.forEach(e -> e.onChangeValue(configuration, key, oldValue, newValue));
}

ypiel-talend: Ok the method exists after...

ypiel-talend: Please check if key exists...

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @param key configuration key
+     * @param newValue new value to set in migration
+     * @param condition predicate that tests current value, if true sets new value
+     */
+    public final void changeValue(final String key, final String newValue, final Predicate<String> condition) {
+        final String oldValue = configuration.get(key);
+        if (condition.test(oldValue)) {
+            configuration.put(key, newValue);
```
ypiel-talend, line 125: You should directly call "changeValue(final String key, final String newValue)"

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * @param key configuration key
+     * @param updater function to set new value
+     */
+    public final void changeValue(final String key, final Function<? super String, String> updater) {
+        final String oldValue = configuration.get(key);
+        final String newValue = updater.apply(oldValue);
+        configuration.put(key, newValue);
```
ypiel-talend, line 138: You should directly call "changeValue(final String key, final String newValue)"

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/component/AbstractMigrationHandler.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param key configuration key
+     * @param updater function to set new value
+     * @param condition predicate that tests current value, if true sets new value
+     */
+    public final void changeValue(final String key, final Function<? super String, String> updater,
+            final Predicate<String> condition) {
+        final String oldValue = configuration.get(key);
+        if (condition.test(oldValue)) {
+            final String newValue = updater.apply(oldValue);
+            configuration.put(key, newValue);
```
ypiel-talend, line 153: You should directly call "changeValue(final String key, final String newValue)"

ypiel-talend (changes requested):
Some remarks

undx:
Made changes but I'm not sure that we should harden so much !
I think that logging an error is perhaps better than blocking with exceptions. We may have many legacy issues...
@ypiel-talend WDT?

ypiel-talend:
> Made changes but I'm not sure that we should harden so much !
> I think that logging an error is perhaps better than blocking with exceptions. We may have many legacy issues...
> @ypiel-talend WDT?

For me log can be too late. If a key was expected and was not or if we add a key but in fact it was already here and so we override it...
I suggested to throw an Exception so then the developer of the migration task can manage himself what to do if a key doesn't exist. SO maybe we should define a new exception like : MigrationException extends Exception and not RuntimeException.
So then, the dev will have to do:
---------------------------------
try{
	this.changeValue("mykey", "new value");
	this.addKey("newkey", "new value");
}
catch(MigrationException e){
	throw new RuntimeException("Cant do migration for xxxxx. The current configuration is not consistent.", e);
}

String toRemoveKey = "toremove";
try{
	this.removeKey(toRemoveKey) {
}
catch(MigrationException e){
	log.warn("The key '"+toRemoveKey+"' to remove was already not set");
}
------------------------------------------------------

I would have defined methods like this :
-----------------------------------------------
public final void addKey(final String key, final String value)  throw MigrationException {...}
-----------------------------------------------

ypiel-talend (changes requested):
One comment

ypiel-talend (dismissed):
LGTM

PR #448, opened 2020-10-14
fix(TCOMP-1722): respect user's url input

undx:
https://jira.talendforge.org/browse/TCOMP-1722

Removing last slash in user input is, IMHO, not a good thing. Having an ending slash has a particular meaning in REST resources so not reflecting the user's initial intent.


ypiel-talend (approved):
LGTM

PR #456, opened 2021-01-04
TCOMP-1811 : primitif to Object

clesaec:
https://jira.talendforge.org/browse/TCOMP-1811

### Requirements

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
@@ -130,10 +130,10 @@ public final class RecordImpl implements Record {
 
         private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type) {
             if (providedSchema == null) {
-                return new SchemaImpl.EntryImpl.BuilderImpl()
-                        .withName(name)
-                        .withType(type)
-                        .withNullable(true)
+                return new SchemaImpl.EntryImpl.BuilderImpl() //
+                        .withName(name) //
+                        .withType(type) //
+                        .withNullable(true) //
                         .build();
             }
             return this.findExistingEntry(name);
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
@@ -128,12 +128,12 @@ public final class RecordImpl implements Record {
             return entry;
         }
 
-        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type, final boolean nullable) {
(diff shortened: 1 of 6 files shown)


undx (changes requested):
This PR lacks of UTs please.

clesaec:
Good remark :+1: 
UT added (and one correction)

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -128,12 +128,12 @@ public BuilderImpl(final Schema providedSchema) {
...
         }
 
-        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type, final boolean nullable) {
+        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type) {
             if (providedSchema == null) {
                 return new SchemaImpl.EntryImpl.BuilderImpl()
                         .withName(name)
                         .withType(type)
-                        .withNullable(nullable)
+                        .withNullable(true)
```
undx, line 136: The goal of the PR is to return Objects and not primitives (so we may have null values indeed) but the fact is that WithNullable is a schema parameter that must be honored when setting/getting the record value by the implementation. Hope that I'm clear otherwise ping me on slack.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -128,12 +128,12 @@ public BuilderImpl(final Schema providedSchema) {
             return entry;
         }
 
-        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type, final boolean nullable) {
+        private Schema.Entry findOrBuildEntry(final String name, final Schema.Type type) {
```
undx, line 131: I think that we must keep this schema parameter...

undx: `component-kitap` module has failing tests coming from your changes, see travis PR build.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: This method is "private" on builder to add field schema from its name.
Nullable false was employed for all primitives types (int, long, boolean ...) as a null value could not be assigned. Now that we have no primitive, null value can be everywhere, so nullable is always true.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: that's a a private method, so no large consequence with signature change.
for component-kitap; i just have to change unit test.

undx (approved):
LGTM

PR #462, opened 2021-02-05
Upgrade netty to v4.1.58.Final and ensure default http testing module is java 11 friendly over ssl

rmannibucau:
### Requirements

https://jira.talendforge.org/browse/TCOMP-1852

No code change, tests already there

No change

### Why this PR is needed?

The netty shade in http testing module uses an old netty version not supporting java 11 (take care netty 5 alpha is "pre" most of the v4 now).

* It enables to disable unsafe usage on java 11 (-Dorg.talend.__shade__.io.netty.noUnsafe=true) and therefore avoid java >= 11 illegal access errors/warnings
* It sets the system property to disable host name verification of java 11 http client (not always sufficient but best effort, worse case user has to set it in its system properties/pom)

### DIFF
--- a/component-runtime-testing/component-runtime-http-junit/pom.xml
+++ b/component-runtime-testing/component-runtime-http-junit/pom.xml
@@ -28,7 +28,7 @@
   <description>JUnit integration to simplify the HTTP API testing/mocking.</description>
 
   <properties>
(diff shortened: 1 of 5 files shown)


undx (approved):
LGTM Thx Romain !

PR #465, opened 2021-03-01
feat(TDI-45543) : Some fixes for crawling

ypiel-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TDI-45543

Some fixes


undx (approved):
LGTM

PR #466, opened 2021-03-10
feat(TCOMP-1681): add props to schema

undx:
https://jira.talendforge.org/browse/TCOMP-1681

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
@@ -117,10 +117,9 @@ public class SchemaImpl implements Schema {
     @AllArgsConstructor
     public static class EntryImpl implements org.talend.sdk.component.api.record.Schema.Entry {
 
-        // add this for some old code which refer this construct
         public EntryImpl(final String name, final Schema.Type type, final boolean nullable, final Object defaultValue,
                 final Schema elementSchema, final String comment, final Map<String, String> props) {
-            this.name = name;
+            initNames(name);
             this.type = type;
             this.nullable = nullable;
             this.defaultValue = defaultValue;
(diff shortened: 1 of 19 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -39,6 +41,19 @@
 
     private List<Entry> entries;
 
+    private Map<String, String> props = new LinkedHashMap<>(0);
+
+    public SchemaImpl(final Type type, final Schema schema, final List<Entry> entries) {
+        this.type = type;
+        elementSchema = schema;
+        this.entries = entries;
+    }
```
clesaec, line 50: Like this, with "@AllArgsConstructor" and "@Data" annotations, SchemaImpl instance become mutable.
```java
new SchemaImpl(x, y, z, myMap); myMap.put(xx, zz);
schema.getProps().put(xx, zz)
```
1 . put explicit constructor : 
```java
public SchemaImpl(final Type type, final Schema schema, final List<Entry> entries, Map<String, String> xx) {
   Map<String, String> val = new ...;
   val.addAll(xx);
   this.props = Collections.unmodifiableMap(val);  // while we don't use [vavr lib](https://docs.vavr.io/) to have real immutable collection.
}
```
2. make props final field

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -47,12 +62,14 @@
 
         private List<Entry> entries = new ArrayList<>();
 
+        private Map<String, String> props = new LinkedHashMap<>(0);
```
clesaec, line 65: While use linkedHashMap here ? really need to keep props order ?

clesaec (commented):
One remark on schemaImpl immutability

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: Initially, the schema is mutable. We don't have to make it immutable, this may break some stuff.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: Yes, this is deliberated! This choice is relevant to Avro schema wrapping, internally `Schema` uses the `LinkedHashMap` class for its props.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
clesaec: You mean some code use "schema.getEntries().add( xxx );" instead of using Schema.Builder class ? (except that, Schema interface leaves no place to mutation)
On SchemaImpl : type, elementSchema and entries are not final, but it should.

undx: You're right but this is not in scope of the changes. Please create a new issue for that.

clesaec: done : https://jira.talendforge.org/browse/TCOMP-1876

PR #467, opened 2021-03-10
feat(TCOMP-705): after variables for connectors

ozhelezniak-talend:
* extract runtime value via reflection
* add checks
* populate data with metadata
* new component metadata enricher

### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-705

Add after variables support in the Studio.

### Related PR:
[Use in studio-integration plugin](https://github.com/Talend/tdi-studio-se/pull/5933)


component-api/src/main/java/org/talend/sdk/component/api/component/AfterVariableContainer.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Mark method that return container with after variables.
+ */
+@Target(ElementType.METHOD)
+@Retention(RUNTIME)
+public @interface AfterVariableContainer {
```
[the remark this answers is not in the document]
undx: Can't we link this notion to `@AfterVariables` annotation with a value name like `method` or smtg like this? So, we can get rid of this annotation...
Retain this comment on hold, coz reading the rest of code...

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/extension/AfterVariableMetadataEnricher.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private static final String VALUE_DELIMITER = "\\:";
+
+    private static final String LINE_DELIMITER = "\\;";
+
+    static {
+        SUPPORTED_ANNOTATIONS.add(PartitionMapper.class);
+        SUPPORTED_ANNOTATIONS.add(Emitter.class);
+        SUPPORTED_ANNOTATIONS.add(Output.class);
+    }
```
undx, line 50: You can init the set using `new HashSet<>(Arrays.asList(PartitionMapper.class, Emitter.class, ...) )` instead of static block.

component-runtime-manager/src/main/resources/META-INF/services/org.talend.sdk.component.spi.component.ComponentMetadataEnricher
```diff
@@ -0,0 +1 @@
+org.talend.sdk.component.runtime.manager.extension.AfterVariableMetadataEnricher
```
undx, line 1: Maybe this spi should put in studio integration part, otherwise it will be in all envs...

undx (commented):
Some small comments.
Nice job Oleksandr! 👍🏻

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/extension/AfterVariableMetadataEnricher.java
ozhelezniak-talend: Nice idea, thank you!

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/visitor/ModelVisitor.java
```diff
@@ -204,4 +213,64 @@ private boolean validOutputParam(final Parameter p) {
     private Stream<Class<? extends Annotation>> getSupportedComponentTypes() {
         return Stream.of(Emitter.class, PartitionMapper.class, Processor.class);
     }
+
+    private void validateAfterVariableContainer(final Class<?> type) {
+        // component can't have more then one after variable container
```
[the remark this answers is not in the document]
ozhelezniak-talend: thx

component-runtime-manager/src/main/resources/META-INF/services/org.talend.sdk.component.spi.component.ComponentMetadataEnricher
ozhelezniak-talend: I agree that it'll be better. But right now not either studio-integration or component-runtime-di jars aren't available at the moment when org.talend.sdk.component.runtime.manager.reflect.ComponentMetadataService#ComponentMetadataService loads spi-s.

component-api/src/main/java/org/talend/sdk/component/api/component/AfterVariableContainer.java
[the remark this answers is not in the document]
ozhelezniak-talend: I'm gonna move this annotation to AfterVariables as an inner class.
I think to keep it. This annotation allows us to change implementation easier if we want to in the future. This annotation just marks something as the container of after variables. And the container may change. For example, we can mark a field, instead of a method, or even use some kind of @Service with it. 
Also, the connector developer isn't forced to use @AfterVariables annotation. It's optional because users can use repeatable @AfterVariable annotations instead.

component-api/src/main/java/org/talend/sdk/component/api/component/AfterVariables.java
```diff
@@ -0,0 +1,58 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * {@link AfterVariable}
+ */
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface AfterVariables {
```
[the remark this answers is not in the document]
ozhelezniak-talend: Added additional explanation to javadoc, marked with @Documentation annotation and section how to use in javajet-to-componentkit.adoc.

component-runtime-manager/src/main/resources/META-INF/services/org.talend.sdk.component.spi.component.ComponentMetadataEnricher
undx: Hum, strange that the SPI is not working in `org.talend.sdk.component.studio-integration`, the component-server is started from there...
Will double check this.

ozhelezniak-talend: I saw that "org.talend.sdk.component.studio-integration" comunicates with "componen-server" via WebSocket.

undx: Yes, using websocket but http is available also. This doesn't interfere in anything and process.
I will create a backlog task for investigating on spi loading in studio

undx (approved):
LGTM

PR #472, opened 2021-03-19
TCOMP-1884 : Add getValue & with method on Record.Builder

clesaec:
https://jira.talendforge.org/browse/TCOMP-1884

Add util method on Record.Builder
Unit tests added on component-api & component-runtime-impl
Correction done on sanitzeConnectionName on component-api (after fails on unit test)

### Requirements

### DIFF
--- a/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
@@ -18,7 +18,10 @@ package org.talend.sdk.component.api.record;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.StandardCharsets;
+import java.time.temporal.Temporal;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -63,16 +66,41 @@ public interface Schema {
     String getProp(String property);
 
     enum Type {
-        RECORD,
-        ARRAY,
-        STRING,
-        BYTES,
-        INT,
-        LONG,
-        FLOAT,
-        DOUBLE,
-        BOOLEAN,
-        DATETIME
+        RECORD(new Class<?>[] { Record.class }),
+        ARRAY(new Class<?>[] { Collection.class }),
+        STRING(new Class<?>[] { String.class }),
+        BYTES(new Class<?>[] { byte[].class, Byte[].class }),
+        INT(new Class<?>[] { Integer.class }),
+        LONG(new Class<?>[] { Long.class }),
+        FLOAT(new Class<?>[] { Float.class }),
+        DOUBLE(new Class<?>[] { Double.class }),
+        BOOLEAN(new Class<?>[] { Boolean.class }),
+        DATETIME(new Class<?>[] { Long.class, Date.class, Temporal.class });
+
+        /** All compatibles Java classes */
+        private final Class<?>[] classes;
+
+        Type(final Class<?>[] classes) {
+            this.classes = classes;
+        }
+
+        /**
+         * Check if input can be affected to an entry of this type.
+         * 
+         * @param input : object.
+         * @return true if input is null or ok.
+         */
+        public boolean isCompatible(final Object input) {
+            if (input == null) {
+                return true;
+            }
+            for (Class<?> clazz : classes) {
+                if (clazz.isInstance(input)) {
+                    return true;
+                }
+            }
+            return false;
+        }
     }
 
     interface Entry {
--- a/component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordBuilderFactory.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordBuilderFactory.java
@@ -15,6 +15,9 @@
  */
 package org.talend.sdk.component.api.service.record;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.talend.sdk.component.api.record.Record;
 import org.talend.sdk.component.api.record.Schema;
 
@@ -62,4 +65,27 @@ public interface RecordBuilderFactory {
      * @return a builder to create a schema entry.
      */
     Schema.Entry.Builder newEntryBuilder();
+
+    /**
+     * Build a schema.entry from another one. Useful to duplicate a column with some changes.
+     * 
+     * @param model : model of entry to copy.
+     * @return entry builder with model parameters.
+     */
+    default Schema.Entry.Builder newEntryBuilder(Schema.Entry model) {
+        final Map<String, String> props = new HashMap<>();
+        final Map<String, String> modelProps = model.getProps();
+        if (modelProps != null) {
+            props.putAll(modelProps);
+        }
+        return this
+                .newEntryBuilder()
+                .withType(model.getType())
+                .withNullable(model.isNullable())
+                .withName(model.getName())
+                .withElementSchema(model.getElementSchema())
+                .withDefaultValue(model.getDefaultValue())
+                .withComment(model.getComment())
+                .withProps(props);
+    }
 }
--- a/component-api/src/test/java/org/talend/sdk/component/api/record/SchemaTest.java
+++ b/component-api/src/test/java/org/talend/sdk/component/api/record/SchemaTest.java
@@ -17,8 +17,11 @@ package org.talend.sdk.component.api.record;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.time.ZonedDateTime;
+import java.time.temporal.Temporal;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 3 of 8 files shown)


component-api/src/main/java/org/talend/sdk/component/api/component/AfterVariables.java
```diff
@@ -33,8 +33,7 @@
  */
 @Documentation("Declare the group of after variable `@AfterVariable`, "
         + "only supported on components - `@PartitionMapper`, `@Processor`, `@Emitter`."
-        + "The functionality is for the Studio only."
-)
```
undx, line 37: please remove anything not related to PR.

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
```diff
@@ -200,21 +231,26 @@ default Entry getEntry(final String name) {
...
-                sanitized[i - offset] = '_';
+            char current = original[i];
+            if ((!ascii.canEncode(current)) || (!Character.isLetterOrDigit(current) && current != '_')) {
+                sanitized[i] = '_';
             } else {
-                sanitized[i - offset] = original[i];
+                sanitized[i] = current;
             }
         }
         return new String(sanitized);
```
undx, line 256: Sorry but please attach this change to a new jira issue. This is really a bug fix and not related to PR but great catch 👍🏻

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
```diff
@@ -60,16 +66,41 @@ default Entry getEntry(final String name) {
...
+        /**
+         * Check if input can be affected to an entry of this type.
+         * 
+         * @param input : object.
+         * @return true if input is null or ok.
+         */
+        public boolean isCompatible(final Object input) {
+            if (input == null) {
+                return true;
+            }
```
undx, line 96: just a remark, we have the `nullable` in entry. checking rest of code...

component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordBuilderFactory.java
```diff
@@ -62,4 +65,27 @@
...
      */
     Schema.Entry.Builder newEntryBuilder();
+
+    /**
+     * Build a schema.entry from another one. Useful to duplicate a column with some changes.
+     * 
+     * @param model : model of entry to copy.
+     * @return entry builder with model parameters.
+     */
+    default Schema.Entry.Builder newEntryBuilder(Schema.Entry model) {
```
undx, line 75: Why not... but don't we have enough?

component-api/src/test/java/org/talend/sdk/component/api/service/interceptor/InterceptorHandlerTest.java
```diff
@@ -0,0 +1,74 @@
...
+        final MyAnnot annot = m1.get();
+        Assertions.assertEquals("C1.m1", annot.xx());
+
+        final Optional<MyAnnot> m2 = handler.findAnnotation(C2.class.getDeclaredMethod("m2"), MyAnnot.class);
+        Assertions.assertNotNull(m2);
+        Assertions.assertTrue(m2.isPresent());
+        final MyAnnot annot2 = m2.get();
+        Assertions.assertEquals("C2.m2", annot2.xx());
+    }
+}
```
undx, line 74: hum, maybe too tired but I don't get the point of this test class related to PR...

component-runtime-beam/src/test/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecordBuilderTest.java
```diff
@@ -73,12 +73,12 @@ void copyRecord() {
...
+                .withRecord("_address",
                         factory.newRecordBuilder(address).withString("street", "here").withInt("number", 1).build())
                 .build();
         final Record output = factory.newRecordBuilder(customSchema, baseRecord).withString("custom", "added").build();
         assertEquals(
-                "AvroRecord{delegate={\"name\": \"Test\", \"age\": 33, \"address\": {\"street\": \"here\", \"number\": 1}, \"custom\": \"added\"}}",
+                "AvroRecord{delegate={\"name\": \"Test\", \"age\": 33, \"_address\": {\"street\": \"here\", \"number\": 1}, \"custom\": \"added\"}}",
                 output.toString());
     }
 }
```
undx, line 84: see my comment upper on splitting PR responsibility....

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -115,6 +118,37 @@ public BuilderImpl(final Schema providedSchema) {
             this.providedSchema = providedSchema;
         }
 
+        @Override
+        public Object getValue(final String name) {
+            return this.values.get(name);
+        }
+
+        @Override
+        public Builder with(final Entry entry, final Object value) {
```
undx, line 127: Why not... but same remark as above.

undx (commented):
I'm not totally sure to understand the goal of this PR...
I don't get the point _Record.Builder instances are currently opaque, you can't see current state (values)._
This is supposed to be when using a builder... so following steps should be enough clear, no?

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
clesaec: Here, it's just to check compatibility with Schema.Type enum (as instanceof), so null is OK

component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordBuilderFactory.java
clesaec: It's a facility function, similar to 
```java
Schema.Builder newSchemaBuilder(Schema schema);
Record.Builder newRecordBuilder(Schema schema, Record record);
```
and can help on DataPrep processor, where we can add a new Field that is a copy of an existing one but with a different name.

component-api/src/test/java/org/talend/sdk/component/api/service/interceptor/InterceptorHandlerTest.java
clesaec: Yes, not related to the PR
I needed to watch how InterceptorHandler worked, add this unit test and as this is the only direct unit test for this class and take few millis to run, i thought it was good to keep it.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
clesaec: May simplify client code like : "if (entry.getType() ... && value instanceof) ..."; only constraint is that type control become dynamic (like for nullable).

component-api/src/test/java/org/talend/sdk/component/api/service/interceptor/InterceptorHandlerTest.java
undx: ok, let's keep it although not related...

component-api/src/test/java/org/talend/sdk/component/api/record/SchemaTest.java
```diff
@@ -135,6 +135,20 @@ public String getProp(String property) {
...
+        Assertions.assertTrue(Schema.Type.INT.isCompatible(123));
+        Assertions.assertFalse(Schema.Type.INT.isCompatible(234L));
+        Assertions.assertFalse(Schema.Type.INT.isCompatible("Hello"));
+
+        Assertions.assertTrue(Type.BYTES.isCompatible("Hello".getBytes()));
+        Assertions.assertTrue(Type.BYTES.isCompatible(new Byte[] {}));
+        Assertions.assertFalse(Schema.Type.BYTES.isCompatible(new int[] {}));
+
+        Assertions.assertTrue(Type.DATETIME.isCompatible(System.currentTimeMillis()));
+    }
```
undx, line 150: extend a bit test to all types....

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
```diff
@@ -63,16 +66,41 @@ default Entry getEntry(final String name) {
...
-        DATETIME
+        RECORD(new Class<?>[] { Record.class }),
+        ARRAY(new Class<?>[] { Collection.class }),
+        STRING(new Class<?>[] { String.class }),
+        BYTES(new Class<?>[] { byte[].class, Byte[].class }),
+        INT(new Class<?>[] { Integer.class }),
+        LONG(new Class<?>[] { Long.class }),
+        FLOAT(new Class<?>[] { Float.class }),
+        DOUBLE(new Class<?>[] { Double.class }),
+        BOOLEAN(new Class<?>[] { Boolean.class }),
```
undx, line 77: Shouldn't we include primitives classes as in MappingUtils? (ie int.class, etc.)

undx (approved):
Some minor comments but LGTM.

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
clesaec: As "public boolean isCompatible(final Object input)" as an Object in input, a call with a primitive will be "boxed" into corresponding object; so primitives type is not useful.

component-api/src/test/java/org/talend/sdk/component/api/record/SchemaTest.java
clesaec: done

PR #475, opened 2021-03-25
feat(TCOMP-1814): Standalone connectors support

chmyga:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-1814

Standalone connectors support and processing in framework. Does not include beam/cloud implementation and processing.


undx (commented):
Apparently, LGTM! Will try to do some tests this week-end for checking, will approve afterwards.

Great task Dmytro ! 👍🏻

undx (approved):
LGTM 👍🏻

ypiel-talend (commented):
LGTM

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -1903,6 +1932,46 @@ public String name() {
...
+                            .supports(org.talend.sdk.component.runtime.standalone.DriverRunner.class)
+                                    ? config -> executeInContainer(plugin,
+                                            () -> context
+                                                    .getOwningExtension()
+                                                    .convert(new ComponentInstanceImpl(doInvoke(
+                                                            constructor, parameterFactory.apply(config)), plugin,
+                                                            component.getName(), name),
+                                                            org.talend.sdk.component.runtime.standalone.DriverRunner.class))
+                                    : config -> new DriverRunnerImpl(this.component.getName(), name, plugin,
+                                            doInvoke(constructor, parameterFactory.apply(config)));
```
clesaec, line 1959: Would be easier to understand with code like : 
```java
final boolean isDriverSupport = context.getOwningExtension() != null && context
                            .getOwningExtension()
                            .supports(org.talend.sdk.component.runtime.standalone.DriverRunner.class);

final Function<Map<String, String>, org.talend.sdk.component.runtime.standalone.DriverRunner> instantiator =
    (Map<String, String> config) -> {
            final Serializable instance = doInvoke(constructor, parameterFactory.apply(config));
            if (isDriverSupport) {
                     final ComponentInstanceImpl cpnInstance = new ComponentInstanceImpl(instance, plugin,
                                                            component.getName(), name);
                     return executeInContainer(plugin,
                               context.getOwningExtension()
                                           .convert(cpnInstance,
                                                            org.talend.sdk.component.runtime.standalone.DriverRunner.class))
              }
              else {
                       new DriverRunnerImpl(this.component.getName(), name, plugin,   instance);
            }
   }
```
but that's my poor brain which is not adapt enough :)

clesaec (approved):
:+1: 
with one comment

PR #476, opened 2021-03-26
use annotation to replace interface for auto inject

wwang-talend:


undx (approved):
That's nicer ;-) 
LGTM

PR #477, opened 2021-03-29
feat(TCOMP-1815): return a proper error payload for migration

undx:
https://jira.talendforge.org/browse/TCOMP-1815


ypiel-talend (approved):
LGTM

PR #485, opened 2021-04-22
feat(TCOMP-1898): Add removeEntry on Record.Builder

pgautier83:
### Requirements


clesaec (approved):
May find a more general method for Builder `updateEntryByName(String name, Schema.Entry schemaEntry);` as `Builder updateEntry(Predicate<Schema.Entry> selector, Schema.Entry schemaEntry);`
But, OK

PR #492, opened 2021-05-31
fix(TCOMP-1925): fix configuration mapper

ozhelezniak-talend:
https://jira.talendforge.org/browse/TCOMP-1925

### Requirements


undx (approved):
LGTM

PR #494, opened 2021-06-01
feat(TDI-45570): add code coverage reporting module

undx:
https://jira.talendforge.org/browse/TDI-45570


ozhelezniak-talend (approved):
Looks ok)

PR #495, opened 2021-06-08
fix(TCOMP-1935) after variables

ozhelezniak-talend:
### Requirements

### Why this PR is needed?
Fix of After Variables Feature.

https://jira.talendforge.org/browse/TCOMP-1935
It adds validation of what types the connector-writer can use when defines After variables.
I think List-type is still a tricky one and maybe a backdoor for us. But at least it's supported by
https://github.com/Talend/studio-tools/blob/12808863466ad7adecc3e24efe9db8c5489fd825/org.talend.cloud.publisher/src/main/java/org/talend/core/model/metadata/types/JavaTypesManager.java


undx (approved):
LGTM

PR #500, opened 2021-06-18
TCOMP-1923 : metadata in records

clesaec:
https://jira.talendforge.org/browse/TCOMP-1923
(need to adapt unit test on some connectors that use directly SchemaImpl instead of RecordBuilderFactory (couhbase, azure, dynamoDB)
### Requirements

### DIFF
--- a/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import javax.json.Json;
 import javax.json.JsonValue;
@@ -45,9 +46,14 @@ public interface Schema {
     Schema getElementSchema();
 
     /**
-     * @return the entries for records.
+     * @return the data entries for records (not contains meta data entries).
      */
     List<Entry> getEntries();
+
+    /**
+     * @return All entries, including data and metadata, of this schema.
+     */
+    Stream<Entry> getAllEntries();
 
     default Entry getEntry(final String name) {
         return Optional
@@ -154,6 +160,11 @@ public interface Schema {
         boolean isNullable();
 
         /**
+         * @return true if this entry is for metadata; false for ordinary data.
+         */
+        boolean isMetadata();
+
+        /**
          * @param <T> the default value type.
          * @return Default value for this entry.
          */
@@ -212,6 +223,8 @@ public interface Schema {
             Builder withType(Type type);
 
             Builder withNullable(boolean nullable);
+
+            Builder withMetadata(boolean metadata);
 
             <T> Builder withDefaultValue(T value);
 
--- a/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
@@ -51,6 +51,11 @@ public interface Schema {
     List<Entry> getEntries();
 
     /**
+     * @return the metadata entries for records (not contains ordinary data entries).
+     */
+    List<Entry> getMetadata();
+
+    /**
      * @return All entries, including data and metadata, of this schema.
      */
     Stream<Entry> getAllEntries();
--- a/component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
+++ b/component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
@@ -132,7 +132,7 @@ class RecordBuilderFactoryTest {
 
         @Override
         public Entry.Builder withMetadata(boolean metadata) {
-            this.builder.withNullable(metadata);
+            this.builder.withMetadata(metadata);
             return this;
         }
 
--- a/component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
+++ b/component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
@@ -78,6 +78,8 @@ class RecordBuilderFactoryTest {
 
         private final boolean nullable;
 
+        private final boolean metadata;
+
         private final Object defaultVal;
 
         @Override
@@ -125,6 +127,12 @@ class RecordBuilderFactoryTest {
         @Override
         public Entry.Builder withNullable(boolean nullable) {
             this.builder.withNullable(nullable);
+            return this;
+        }
+
+        @Override
+        public Entry.Builder withMetadata(boolean metadata) {
+            this.builder.withNullable(metadata);
             return this;
         }
 
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
@@ -36,9 +36,11 @@ import org.talend.sdk.component.runtime.manager.service.api.Unwrappable;
 import org.talend.sdk.component.runtime.record.SchemaImpl.EntryImpl;
 
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 import lombok.ToString;
 
 @Data
+@EqualsAndHashCode(of = "delegate")
 @ToString(of = "delegate")
 public class AvroSchema implements org.talend.sdk.component.api.record.Schema, AvroPropertyMapper, Unwrappable {
 
@@ -112,11 +114,13 @@ public class AvroSchema implements org.talend.sdk.component.api.record.Schema, A
     }
 
     @Override
-    @JsonbTransient
-    public Stream<Entry> getAllEntries() {
+    public List<Entry> getMetadata() {
+        if (getActualDelegate().getType() != Schema.Type.RECORD) {
+            return emptyList();
+        }
         if (this.metadataEntries == null) {
             synchronized (this) {
-                if (entries != null) {
+                if (this.metadataEntries == null) {
                     this.metadataEntries = this
                             .getNonNullFields() //
                             .filter(AvroSchema::isMetadata) // only metadata fields
@@ -125,8 +129,13 @@ public class AvroSchema implements org.talend.sdk.component.api.record.Schema, A
                 }
             }
         }
+        return this.metadataEntries;
+    }
 
-        return Stream.concat(this.getEntries().stream(), this.metadataEntries.stream());
+    @Override
+    @JsonbTransient
+    public Stream<Entry> getAllEntries() {
+        return Stream.concat(this.getEntries().stream(), this.getMetadata().stream());
     }
 
     private Stream<Field> getNonNullFields() {
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
@@ -24,13 +24,16 @@ import static org.talend.sdk.component.runtime.beam.avro.AvroSchemas.unwrapUnion
 
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.json.bind.annotation.JsonbTransient;
 
 import org.apache.avro.LogicalTypes;
 import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field;
 import org.talend.sdk.component.runtime.manager.service.api.Unwrappable;
-import org.talend.sdk.component.runtime.record.SchemaImpl;
+import org.talend.sdk.component.runtime.record.SchemaImpl.EntryImpl;
 
 import lombok.Data;
 import lombok.ToString;
@@ -39,12 +42,18 @@ import lombok.ToString;
 @ToString(of = "delegate")
 public class AvroSchema implements org.talend.sdk.component.api.record.Schema, AvroPropertyMapper, Unwrappable {
 
+    // alias that indicate field is metadata if present.
+    static final String METADATA_ALIAS_NAME = "__METADATA__";
+
     @JsonbTransient
     private final Schema delegate;
 
     private volatile AvroSchema elementSchema;
 
     private volatile List<Entry> entries;
+
+    @JsonbTransient
+    private volatile List<Entry> metadataEntries;
 
     private volatile Type type;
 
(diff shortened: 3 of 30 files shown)


component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordService.java
```diff
@@ -86,6 +86,7 @@ Record create(Schema schema, Record fallbackRecord, BiFunction<Schema.Entry, Rec
      * Visit a record with a custom visitor.
      *
      * @param visitor the visitor to use to browse the record.
+     * @param record recird to visit.
```
undx, line 89: typo

component-api/src/main/java/org/talend/sdk/component/api/service/http/Configurer.java
```diff
@@ -81,6 +81,8 @@
 
         /**
          * Prevents the client to follow redirections.
+         * 
+         * @return connection.
```
undx, line 85: not in scope...

component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
```diff
@@ -128,6 +130,12 @@ public String getProp(final String property) {
             return this;
         }
 
+        @Override
+        public Entry.Builder withMetadata(boolean metadata) {
+            this.builder.withNullable(metadata);
```
undx, line 135: Not sure it's intended!

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
```diff
@@ -39,13 +42,19 @@
 @ToString(of = "delegate")
 public class AvroSchema implements org.talend.sdk.component.api.record.Schema, AvroPropertyMapper, Unwrappable {
 
+    // alias that indicate field is metadata if present.
+    static final String METADATA_ALIAS_NAME = "__METADATA__";
```
undx, line 46: Use the already existing `KeysForAvroProperty` and prefix this constant with a `talend.xxx` namespace.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
```diff
@@ -96,21 +105,63 @@ public Type getType() {
...
+
+        return Stream.concat(this.getEntries().stream(), this.metadataEntries.stream());
+    }
+
+    private Stream<Field> getNonNullFields() {
+        return getActualDelegate().getFields().stream().filter(it -> it.schema().getType() != NULL);
+    }
+
+    private static boolean isMetadata(final Field f) {
+        return f.aliases() != null && f.aliases().contains(AvroSchema.METADATA_ALIAS_NAME);
```
undx, line 140: I don't know the usage of `aliases`, what's the purpose?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -22,46 +22,62 @@
...
 
     @Override
     public String getProp(final String property) {
         return props.get(property);
     }
 
+    @Override
+    @JsonbTransient
+    public Stream<Entry> getAllEntries() {
+        return Stream.concat(this.metadataEntries.stream(), this.entries.stream());
```
undx, line 69: here we may loose the ordering, no?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -22,46 +22,62 @@
...
     public static class BuilderImpl implements Builder {
 
         private Type type;
 
         private Schema elementSchema;
 
-        private List<Entry> entries = new ArrayList<>();
+        private final List<Entry> entries = new ArrayList<>();
+
+        private final List<Entry> metadataEntries = new ArrayList<>();
```
undx, line 80: Why use 2 lists?

undx (commented):
AFAIR, we didn't talked about managing two lists of entries.
We also talked about using predicates for getting entries, see slack discussion.
I think there's not enough tests on the features.

component-api/src/main/java/org/talend/sdk/component/api/service/record/RecordService.java
clesaec: Fixed

component-api/src/main/java/org/talend/sdk/component/api/service/http/Configurer.java
clesaec: Yes, but generate build issue without.

component-api/src/test/java/org/talend/sdk/component/api/service/record/RecordBuilderFactoryTest.java
clesaec: Fixed

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
clesaec: I didn't know this interface.
Done

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
clesaec: This is how we store the fact that a field is a metadata field in apache Avro record.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
clesaec: No;
Collection.stream() method transform collection to stream with boolean parallel parameter to false.
So, concat will also have parallel to false; so order will be preserved.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
clesaec: The method "List<Entry> getEntries();" of Schema interface already exist and is widely used by our connectors.
Using one list would lead to rewrite this method with "entries.stream().filter( !metadata).collect(Collectors.toList())" and so reconstruct a new List and degrade performance.

component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java
```diff
@@ -45,10 +46,15 @@
...
     /**
-     * @return the entries for records.
+     * @return the data entries for records (not contains meta data entries).
      */
     List<Entry> getEntries();
 
+    /**
+     * @return All entries, including data and metadata, of this schema.
+     */
+    Stream<Entry> getAllEntries();
```
undx, line 56: Provide also a `getMetadata()` method that returns a `List<Entry>` as for entries.

undx (commented):
In my previous review: 

> AFAIR, we didn't talked about managing two lists of entries.
> We also talked about using predicates for getting entries, see slack discussion.
> **I think there's not enough tests on the features.**

Did you tried to make such kind of test?
```   
@Test
void testRecordWithMetadataFields() {
    final Schema schema = new BuilderImpl() //
        .withType(Type.RECORD) //
        .withEntry(newEntry("field1", "field1", Type.STRING, true, 5, "Comment"))
        .withEntry(new EntryImpl.BuilderImpl().withName("record_id").withMetadata(true).withType(Type.INT).withProp("method","FIFO").build())
        .withEntry(new EntryImpl.BuilderImpl().withName("field2").withMetadata(true).withType(Type.STRING).build())
        .build();
    final RecordImpl.BuilderImpl builder = new RecordImpl.BuilderImpl(schema);
    Record record = builder //
        .withInt("record_id", 34) //
        .withString("field1", "Aloa") //
        .withString("field2", "Hallo, wie gehst du ?") //
        .build();
    Schema recordSchema = record.getSchema();
    assertEquals(1, recordSchema.getEntries().size());
    assertEquals(2, recordSchema.getAllEntries().filter(e-> e.isMetadata()).collect(Collectors.toList()).size());
    assertEquals(34, record.getInt("record_id"));
    assertEquals("Aloa", record.getString("field1"));
    assertEquals("Hallo, wie gehst du ?", record.getString("field2"));
} 
```

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: With the following code, I created some entries in a certain order...
```
final Schema schema = new BuilderImpl() //
                .withType(Type.RECORD) //
                .withEntry(newEntry("field1", "field1", Type.STRING, true, 5, "Comment"))
                .withEntry(new EntryImpl.BuilderImpl().withName("record_id").withMetadata(true).withType(Type.INT).withProp("method","FIFO").build())
                .withEntry(new EntryImpl.BuilderImpl().withName("field2").withMetadata(true).withType(Type.STRING).build())
                .build();
        final RecordImpl.BuilderImpl builder = new RecordImpl.BuilderImpl(schema);
        schema.getAllEntries().forEach(e-> System.out.println(e));
``` 
which outputs
 ```
SchemaImpl.EntryImpl(name=record_id, rawName=null, type=INT, nullable=false, metadata=true, defaultValue=null, elementSchema=null, comment=null, props={method=FIFO})
SchemaImpl.EntryImpl(name=field2, rawName=null, type=STRING, nullable=false, metadata=true, defaultValue=null, elementSchema=null, comment=null, props={})
SchemaImpl.EntryImpl(name=field1, rawName=field1, type=STRING, nullable=true, metadata=false, defaultValue=5, elementSchema=null, comment=Comment, props={})
```
So, order is not preserved.

Comment (clesaec) [reply component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java:69]: yes, metadata fields come before ... then classic field. 
Order is preserved inside it.
Is global ordering important ?

File: component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java (hunk 8 above)
Comment (clesaec) [reply component-api/src/main/java/org/talend/sdk/component/api/record/Schema.java:56]: ok

File: component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordConvertersTest.java [hunk 9]
```diff
@@ -344,6 +345,30 @@ void convertListObject(final JsonBuilderFactory jsonBuilderFactory, final JsonPr
...
+                                    .add("list", Json
+                                            .createArrayBuilder()
+                                            .add(Json.createObjectBuilder().add("name", "a").add("name1", "a1").build())
+                                            .add(Json.createObjectBuilder().add("name", "b").add("name2", "b2").build())
+                                            .build())
+                                    .build(),
+                            () -> jsonb, () -> new RecordBuilderFactoryImpl("test"));
+            final Collection<Record> list = record.getArray(Record.class, "list");
+            final Schema schema = record.getSchema().getEntries().get(0).getElementSchema();
+            // // FIXME: 7/9/21 : record -> schema -> entry(list) -> elementSchema != record -> list[x] -> schema
```
undx, line 365: if there's a potential fix, please create a jira issue and link it's preferable. thanks

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchemaBuilder.java
```diff
@@ -232,11 +224,33 @@ public Schema build() {
             if (elementSchema == null) {
                 throw new IllegalStateException("No elementSchema set for this ARRAY schema");
             }
+            // FIXME: 7/12/21 => Here, it assumes that elementSchema inherit from Unwrappable<avro.Schema>
```
undx, line 227: if there's a potential fix, please create a jira issue and link it's preferable. thanks

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: Thinking so! 
Sorry, I stick to my initial idea... metadata should be treated as a property in schema like the nullable one.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: see my comment below... Schema is not handled on each record (in most cases...) so performance should not be an issue in that case...

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
```diff
@@ -46,6 +51,9 @@
 
     private volatile List<Entry> entries;
 
+    @JsonbTransient
+    private volatile List<Entry> metadataEntries;
```
undx, line 55: see my other comments.

undx (commented):
see my comments. Did not review whole changes since last review...
Also, still missing some UTs for me... 
Maybe, let's take some time to discuss about it tomorrow ;-)

component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordConvertersTest.java
clesaec: https://jira.talendforge.org/browse/TCOMP-1956

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchemaBuilder.java
clesaec: https://jira.talendforge.org/browse/TCOMP-1957

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
clesaec: ?? each record has a schema, [see RecordImpl.BuilderImpl.build()](https://github.com/Talend/component-runtime/blob/master/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java#L273-L274) method.
So, even if a record has no metadata; puting list of data and meta data together will oblige to build a new list on each call of getEntries() method (this.entries.stream().filter(not meta)...) because most of component must not see metadata (it must be transparent until you asked explicitly for it)). So; performance issue would be always here.

Moreover, having a field name "entries" and a method "getEntries" that doest not reflect the list content is perturbing; or, at least, with this solution, i can rename it to allEntries; but performance issue will be here for ALL records even those without metadata.

undx:
After discussing on the main issue of this PR IRL, we agreed to follow suggested implementation. (no impact on perf and feat is really specific...)
Just remains increasing unit testing on feature (mainly on record types).

clesaec:
unit test added.

undx (dismissed):
LGTM
(still my comment on order, we'll see)

undx (dismissed):
LGTM

undx (approved):
LGTM

PR #505, opened 2021-08-05
Security chore(): Switch SAST Scan from sandbox to dedicated Application

Gby56:
We've now got more licenses, and microservices can use their own application, so you don't loose work like a sandbox when it expires !
If you need dedicated branch scanning, use the sandboxes for that instead.

**What is the problem this PR is trying to solve?**
SAST scans are performed in a sandbox that expires and loses work.

**What is the chosen solution to this problem?**
SAST scans are now automatically creating an application, based on the previous sandbox name.

**Please check if the PR fulfills these requirements**
- [X] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

### Requirements

### DIFF
--- a/ci/Jenkinsfile-scan
+++ b/ci/Jenkinsfile-scan
@@ -92,16 +92,16 @@ spec:
                 container('main') {
                     withCredentials([veracodeCredentials, nexusCredentials]) {
                         sh "mvn -B -s .jenkins/settings.xml package $SKIP_OPTS"
-                        veracode applicationName: "$VERACODE_APP_NAME",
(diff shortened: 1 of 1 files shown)


undx (approved):
LGTM

undx (approved):
LGTM

PR #518, opened 2021-10-28
fix(TCOMP-2021): Missing logic when handling null date values in Avro Record

jad-darrous:
https://jira.talendforge.org/browse/TCOMP-2021

### Requirements

### Why this PR is needed?
When a date is `null`, it's stored as `-1` (for unknown reason to me, even though the API support null fields). When it's retrieved back, the `-1` is converted back to `null`. However, this check should be done in all the implementations, and it's not done when converting to Avro IndexedRecord. 

Store null dates as null.

**NB:** This PR is NOT ready for merge. It presents two ways to solve the issue, with some clarifications in the code. Waiting for an expert opinion before polishing the code.


undx (approved):
LGTM

PR #520, opened 2021-11-16
TCOMP-2018 : optimized Avro record conversion from TCK

clesaec:
https://jira.talendforge.org/browse/TCOMP-2018
Optimize conversion of TCK Record to Avro Record by keep a cache of Schema conversion.
(Because, most of time, Records of a pipeline shared same schema)
### Requirements

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
@@ -30,11 +30,9 @@ import javax.json.bind.annotation.JsonbTransient;
 
 import org.talend.sdk.component.api.record.Schema;
 
-import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 
-@EqualsAndHashCode
 @ToString
 public class SchemaImpl implements Schema {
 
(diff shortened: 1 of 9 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -67,6 +65,48 @@
...
+        }
+        return Objects.equals(this.type, other.type)
+                && Objects.equals(this.elementSchema, other.elementSchema)
+                && Objects.equals(this.entries, other.entries)
+                && Objects.equals(this.metadataEntries, other.metadataEntries)
+                && Objects.equals(this.props, other.props);
+    }
+
+    protected boolean canEqual(final SchemaImpl other) {
+        return true;
```
undx, line 107: Don't get the purpose w/ upper usage...

clesaec: It's special for good equals method, see https://www.artima.com/articles/how-to-write-an-equality-method-in-java for example.
Lombok generate a 
```java
protected boolean canEqual(final Object other) {
    return other instanceof SchemaImpl;
}
```
and used by equals as : 
```
if (other instanceof SchemaImpl) {
   if (! (SchemaImpl.canEqual(this)) { return false; }
   ...
}
```
(but as it's used with this that is at least SchemaImpl, i rewrite)
On SchemaImpl subclass, you rewrite as : 
```java
protected boolean canEqual(final SchemaImpl other) {
    return other instanceof SchemaImplSubClass;
}
```

undx: ok got it.

undx (approved):
LGTM

PR #541, opened 2022-01-10
fix(TCOMP-2071): Define default methods on API interfaces

undx:
https://jira.talendforge.org/browse/TCOMP-2071


ozhelezniak-talend (approved):
lgtm

PR #542, opened 2022-01-10
feat(TCOMP-2045): Pass meta information about columns

undx:
https://jira.talendforge.org/browse/TCOMP-2045

*Related PRs*


component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitor.java
```diff
@@ -397,14 +440,31 @@ private Entry toEntry(final String name, final Schema.Type type) {
     }
```
ozhelezniak-talend, line 440: toEntry(final String name, final Schema.Type type)
Is it used?

ozhelezniak-talend (approved):
LGTM

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitor.java
undx: good catch, I missed it 👍🏻

PR #544, opened 2022-02-07
Clesaec/tcomp 2080 avro record fix

clesaec:
https://jira.talendforge.org/browse/TCOMP-2080

### Requirements

### DIFF
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecord.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecord.java
@@ -81,7 +81,7 @@ public class AvroRecord implements Record, AvroPropertyMapper, Unwrappable {
 
                             if (avroValue != null) {
                                 final org.apache.avro.Schema.Field field =
-                                        this.schema.getDelegate().getField(sanitizeConnectionName(entry.getName()));
+                                        this.schema.getActualDelegate().getField(sanitizeConnectionName(entry.getName()));
                                 delegate.put(field.pos(), avroValue);
                             }
                         }));
--- a/component-runtime-beam/src/test/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecordTest.java
+++ b/component-runtime-beam/src/test/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecordTest.java
@@ -203,6 +203,7 @@ class AvroRecordTest {
                     .build();
 
             Assertions.assertNotNull(record);
+            Assertions.assertEquals("world", record.getString("field"));
         }
         finally {
             if (oldValue == null) {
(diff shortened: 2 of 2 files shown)


component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecord.java
```diff
@@ -81,7 +81,7 @@ public AvroRecord(final Record record) {
 
                             if (avroValue != null) {
                                 final org.apache.avro.Schema.Field field =
-                                        this.schema.getDelegate().getField(sanitizeConnectionName(entry.getName()));
+                                        this.schema.getActualDelegate().getField(sanitizeConnectionName(entry.getName()));
```
ypiel-talend, line 84: Should not we also put raw/original name in the according avro property ?

ypiel-talend (commented):
one question

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecord.java
clesaec: Here, we don't create a field, we get an existing one, so original name should have been already saved.
Here, "normally"; entry.getName is already sanitized, and so sanitizeConnectionName may be useless, but generate no bug; so it's for another JIRA.

undx: Apparently, we're also loosing all properties...

clesaec: Fixed for schema properties (and tested for both)

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchemaBuilder.java
```diff
@@ -339,6 +339,7 @@ public Schema build() {
                     .createRecord(SchemaIdGenerator.generateRecordName(avroFields), null, "talend.component.schema",
                             false);
             record.setFields(avroFields);
+            this.props.forEach(record::addProp);
             record.addProp(ENTRIES_ORDER_PROP, fields.stream().map(e -> e.getName()).collect(Collectors.joining(",")));
```
undx, line 343: I suggest this:
```java
record.addProp(ENTRIES_ORDER_PROP, fields.stream().map(e -> e.getName()).collect(Collectors.joining(",")));
props.entrySet()
         .stream()
         .filter(k -> !k.getKey().equals(ENTRIES_ORDER_PROP))
         .forEach(e -> record.addProp(e.getKey(), e.getValue()));
```
if `ENTRIES_ORDER_PROP` was inherited to avoid recdeclaring properties. (will throw an exception).

clesaec: done

undx (commented):
Left an UT proposal on slack

clesaec:
Unit test added

undx (approved):
LGTM

PR #546, opened 2022-02-08
fix(TCOMP-2082): ComponentManager's findDefaultM2 method takes comment as granted

undx:
https://jira.talendforge.org/browse/TCOMP-2082

### DIFF
--- a/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
+++ b/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
@@ -81,6 +81,8 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.jar.JarInputStream;
 import java.util.logging.Level;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -594,7 +596,16 @@ public class ComponentManager implements AutoCloseable {
                 .map(m2 -> {
                     // jobServer may badly translate paths on Windows
                     try {
-                        return PathFactory.get(m2);
+                        log.debug("[findM2] talend.component.manager.m2.repository={}", m2);
+                        final Path m2Path = PathFactory.get(m2);
+                        if (java.nio.file.Files.exists(m2Path)) {
+                            return m2Path;
+                        } else {
+                            log.warn(
+                                    "[findM2] talend.component.manager.m2.repository value points to a non existent path: {}",
+                                    m2Path);
+                            return findStudioM2();
+                        }
                     } catch (Exception e) {
                         return findStudioM2();
                     }
@@ -612,6 +623,7 @@ public class ComponentManager implements AutoCloseable {
                 return localM2;
             }
         }
+        log.debug("[findStudioM2] could not get studio config");
         return findDefaultM2();
     }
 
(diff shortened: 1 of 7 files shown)


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -621,26 +633,38 @@ private static Path findDefaultM2() {
...
-                final int start = content.indexOf("<localRepository>");
                 String localM2RepositoryFromSettings = null;
-                if (start > 0) {
-                    final int end = content.indexOf("</localRepository>", start);
-                    if (end > 0) {
-                        localM2RepositoryFromSettings = content.substring(start + "<localRepository>".length(), end);
+                // do not introduce a xml parser so will do with what we have...
+                final String content = new String(java.nio.file.Files.readAllBytes(settings), StandardCharsets.UTF_8);
+                final Pattern comments = Pattern.compile("(<!--.*?-->)", Pattern.DOTALL);
+                final Pattern emptyLines = Pattern.compile("^\\s*$|\\n|\\r\\n");
```
ozhelezniak-talend, line 643: not an expert here
But maybe we want to include \\r too. What do you think? (Idea says that \\r is line separator in Classic Macos that's why I think it may exist without following \\n)

ozhelezniak-talend: Also, I tried to check it on https://regex101.com/ (mode \gm)
And it doesn't match whitespaces

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/ComponentManagerTest.java
```diff
@@ -76,6 +78,10 @@
 
     private final PluginGenerator pluginGenerator = new PluginGenerator();
 
+    public static final String M2_PROP = "talend.component.manager.m2.repository";
+
+    public static final String SETTINGS_PROP = "talend.component.manager.m2.settings";
```
ozhelezniak-talend, line 83: Didn't know that we have this one.
Also can't find it in documentation

component-runtime-manager/src/test/resources/settings/settings-ok-spaced-mixed-case.xml
```diff
@@ -0,0 +1,269 @@
...
+<settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 http://maven.apache.org/xsd/settings-1.2.0.xsd">
+    <!-- localRepository
+     | The path to the local repository maven will use to store artifacts.
+     |
+     | Default: ${user.home}/.m2/repository
+    <localRepository>/path/to/local/repo</localRepository>
+    -->
+    <lOcAlrEpOsItOrY>
```
ozhelezniak-talend, line 55: nice)

ozhelezniak-talend (approved):
Great job, Emmanuel!
Thank you!
LGTM with minor comments

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
undx: We may add the `\r` but I doubt  that anyone is still using MacOS classic with a ppc :-D 
The goal of `enptyLines` is to flatten file to one big line like this 
```xml
<?xml version="1.0" encoding="UTF-8"?><settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"    xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 http://maven.apache.org/xsd/settings-1.2.0.xsd">        <localRepository>/home/zorro80</localRepository>                <pluginGroups>            </pluginGroups>        <proxies>            </proxies>        <servers>                    </servers>        <mirrors>                <mirror>            <id>maven-default-http-blocker</id>            <mirrorOf>external:http:*</mirrorOf>            <name>Pseudo repository to mirror external repositories initially using HTTP.</name>            <url>http://0.0.0.0/</url>            <blocked>true</blocked>        </mirror>    </mirrors>        <profiles>                    </profiles>    </settings>
```


Its goal is not to reduce extra space. It's done w/ the trim() call on successful capture. 
`^\\s*$` is to match empty line containing 0 or more spaces.

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/ComponentManagerTest.java
undx: yea, you're right it's undocumented... (will fix that later)

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
ozhelezniak-talend: Oh, I see, thank you!
(missed $ at the end of the first part)

PR #548, opened 2022-02-16
fix(TCOMP-1803): add guard to withRecord(string, record)

undx:
https://jira.talendforge.org/browse/TCOMP-1803

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
@@ -427,6 +427,9 @@ public final class RecordImpl implements Record {
         }
 
         public Builder withRecord(final String name, final Record value) {
+            if (value == null) {
+                throw new IllegalArgumentException("No schema for the nested record due to null record value");
+            }
             return withRecord(new SchemaImpl.EntryImpl.BuilderImpl()
                     .withName(name)
                     .withElementSchema(value.getSchema())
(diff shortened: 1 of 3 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -427,6 +427,9 @@ public Builder withRecord(final Schema.Entry entry, final Record value) {
         }
 
         public Builder withRecord(final String name, final Record value) {
+            if (value == null) {
+                throw new IllegalArgumentException("No schema for the nested record due to null record value");
+            }
             return withRecord(new SchemaImpl.EntryImpl.BuilderImpl()
                     .withName(name)
                     .withElementSchema(value.getSchema())
```
ypiel-talend, line 435: what about 
`.withElementSchema(value == null ? null : value.getSchema())` ?
What would be consequences to have a null element schema ?

ypiel-talend (commented):
One question

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
undx: This method cannot handle such case because we need the elementSchema value. If we allow null elementSchema, we will introduce an invalid schema. For a simple use-case, imagine that we use the record's schema for building another one, We have also the case if the RecordImpl has to be wrapped to AvroRecord it will fail.
On the other hand the method withRecord(final Entry entry, final Record value) will work if we provide an elementSchema  (elementSchema not nullable is also checked) and a null Record value.

ypiel-talend (approved):
LGTM

PR #554, opened 2022-03-16
fix(TDI-2102): Fix cloud components image build

quilicicf:
### Requirements

This is just a PR to start a discussion. Not formalized yet.

### Why this PR is needed?

The build of the connectors Docker image fails with SNAPSHOTS today.

What happens is that the talend-component-plugin resolves the jars, then tries to find the associated cars and then extracts the jar from the car.

The issue is that if the cars are not added as dependencies in the pom, then:

* the first time, the plugin will notice that the car is not there and resolve it, which will result in a download from Maven
* then next time it'll find a car with the right `X.Y.Z-SNAPSHOT` in the local repository and believe it's the real deal. The issue is that if in the meantime a new version of the snapshot has been pushed to Nexus, the pom resolution will fetch the new jar but not the new car and the plugin resolution will believe it has the last car when in fact it has an old one. Since we then proceed to get the jar from the car, the jar is out-dated :scream:

The whole logic behind the mojo `prepare-repository` seems to be built on the premise that the single source of truth is the car. It makes no sense to find the jar, assume there's an associated car without checking it's in the dependencies and then get the jar from the car.

The code of the mojo is too complex for me to be comfortable changing it without breaking it. I suggest creating a new mojo to replace the old one and deprecate the old one.

This PR is an attempt at rationalizing the said mojo.

### DIFF
--- a/talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
+++ b/talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
@@ -1,0 +1,291 @@
+/**
+ * Copyright (C) 2006-2022 Talend Inc. - www.talend.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.talend.sdk.component.maven;
+
+import static java.nio.file.Files.copy;
+import static java.nio.file.Files.createDirectories;
+import static java.nio.file.Files.exists;
+import static java.nio.file.Files.newBufferedWriter;
+import static java.nio.file.Files.newInputStream;
+import static java.nio.file.Files.setLastModifiedTime;
+import static java.nio.file.Files.walkFileTree;
+import static java.nio.file.attribute.FileTime.from;
+import static java.time.Instant.ofEpochMilli;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toSet;
+import static org.apache.commons.io.FileUtils.deleteDirectory;
+import static org.apache.commons.io.output.NullOutputStream.NULL_OUTPUT_STREAM;
+import static org.apache.maven.plugins.annotations.LifecyclePhase.PACKAGE;
+import static org.apache.maven.plugins.annotations.ResolutionScope.TEST;
+import static org.talend.sdk.component.maven.api.Audience.Type.TALEND_INTERNAL;
+
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Writer;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.security.DigestOutputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Properties;
+import java.util.Set;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.talend.sdk.component.maven.api.Audience;
+
+@Audience(TALEND_INTERNAL)
+@Mojo(name = "build-repository", defaultPhase = PACKAGE, threadSafe = true, requiresDependencyResolution = TEST)
+public class BuildComponentM2RepositoryMojoV2 extends CarConsumer {
+
+    private static final char[] HEX_CHARS = "0123456789abcdef".toCharArray();
+
+    @Parameter(defaultValue = "true", property = "talend.connectors.write")
+    private Boolean writeVersion;
+
+    @Parameter(defaultValue = "${project.version}", property = "talend.connectors.version")
+    private String version;
+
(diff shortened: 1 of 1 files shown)


talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+    public void doExecute() throws MojoExecutionException {
+        final Set<Artifact> componentCars = project.getDependencies()
+                .stream()
+                .filter(dependency -> "car".equals(dependency.getType())) // FIXME: car should be extracted
+                .filter(dependency -> "compile".equals(dependency.getScope()))
+                .map(dependency -> new DefaultArtifact(
+                        dependency.getGroupId(),
+                        dependency.getArtifactId(),
+                        dependency.getClassifier(),
+                        "car",
```
undx, line 106: could use `dependency.getScope()` as it's filtered above instead of "car". Or follow your FIXME ("car" used 3 times).

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+                        dependency.getVersion()))
+                .map(artifact -> resolve(artifact, classifier, "car")) // No resolve, no file
```
undx, line 108: We may still have issues w/ _-SNAPSHOT_ as `resolve` looks up first with LocalRepositoryManager.

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        componentCars.forEach(this::copyComponentDependencies);
+
+        if (componentCars.isEmpty()) {
+            getLog().error(
+                    "No components found, check the component cars are included in your dependencies with scope compile");
+            throw new IllegalStateException();
```
undx, line 130: throw w/ above message?

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.talend.sdk.component.maven.api.Audience;
+
+@Audience(TALEND_INTERNAL)
+@Mojo(name = "build-repository", defaultPhase = PACKAGE, threadSafe = true, requiresDependencyResolution = TEST)
+public class BuildComponentM2RepositoryMojoV2 extends CarConsumer {
```
undx, line 63: With the changes done, we do not need this inheritance now, so can remove level by 1.
Also, changes should be applied to upper classes (uispec mainly as image mojo is deprecated).

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.talend.sdk.component.maven.api.Audience;
+
+@Audience(TALEND_INTERNAL)
+@Mojo(name = "build-repository", defaultPhase = PACKAGE, threadSafe = true, requiresDependencyResolution = TEST)
```
undx, line 62: Should keep the same goal name. See my comment below on dependency resolution.

undx (commented):
Nice rewriting 👍🏻 
It would have been easier to review w/ changes made directly to targeted class.
Also, see my comments.

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
```diff
@@ -0,0 +1,291 @@
...
+
+    @Parameter(defaultValue = "SHA-512", property = "talend.repository.digestAlgorithm")
+    private String digestAlgorithm;
+
+    @Override
+    public void doExecute() throws MojoExecutionException {
+        final Set<Artifact> componentCars = project.getDependencies()
+                .stream()
+                .filter(dependency -> "car".equals(dependency.getType())) // FIXME: car should be extracted
+                .filter(dependency -> "compile".equals(dependency.getScope()))
```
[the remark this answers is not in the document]
quilicicf: I'd like to discuss this face-to-face to make sure I understand it completely :thinking:

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
quilicicf: Indeed. 
I'm very surprised I couldn't find any constant in the repository for this as it is an extension created by TCK :thinking:

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
quilicicf: There were no issues previously with jars when we used them, were there?
Since we systematically run this mojo with `--update-snapshots` of course which I think is a dangerous gotcha.
Should I change to `resolve(Artifact)` to remove the gotcha? Seems reasonable to me.

```suggestion
                .map(artifact -> resolve(artifact) // Downloads the file if it is not in local repository
```

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/BuildComponentM2RepositoryMojoV2.java
quilicicf: I'm not sure I understand this comment. 
Should I extend `ComponentDependenciesBase` instead of `CarConsumer`?
What changes should be applied to upper classes? :thinking:

quilicicf:
Will re-do this PR now that I don't need a fork anymore.

PR #556, opened 2022-03-23
fix(TDI-2102): Fix cloud components image build

quilicicf:
### Requirements

This code only changes a mojo that is used in a single repository. 

It's been tested for that particular use-case and adding UTs would be overkill at this point.

### Why this PR is needed?

The build of the connectors Docker image fails with SNAPSHOTS today.

What happens is that the `talend-component-plugin` resolves the jars, then tries to find the associated cars and then extracts the jar from the car.

The issue is that if the cars are not added as dependencies in the pom, then:

* the first time, the plugin will notice that the car is not there and resolve it, which will result in a download from Maven
* then next time it'll find a car with the right `X.Y.Z-SNAPSHOT` in the local repository and believe it's the real deal. The issue is that if in the meantime a new version of the snapshot has been pushed to Nexus, the pom resolution will fetch the new jar but not the new car and the plugin resolution will believe it has the last car when in fact it has an old one. Since we then proceed to get the jar from the car, the jar is out-dated :scream:

The whole logic behind the mojo `prepare-repository` seems to be built on the premise that the single source of truth is the car. It makes no sense to find the jar, assume there's an associated car without checking it's in the dependencies and then get the jar from the car.

The code of the mojo is rather complex, I re-wrote it since it's used in a single place and we can easily test that it works. 

* The new mojo searches for car dependencies in the dependencies of the talend-maven-plugin instead of the dependencies of the project (which makes sense since they are used by the plugin). It searches for the cars only, and with compile scope, the test scope was a hack.
* The `ImageM2Mojo` was not used anymore, I removed it.
* Also, I started extracting constants such as the `car` extension.
* Finally, I started documenting things so it's easier to maintain.

### To validate

I still need to validate whether this PR fixes the issue only when using `--update-snapshots` (which we currently use) or not. 

> Note: this is a re-do of https://github.com/Talend/component-runtime/pull/554


talend-component-maven-plugin/pom.xml
```diff
@@ -14,200 +14,200 @@
...
+
+    <reporting>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-plugin-plugin</artifactId>
+                <version>3.5.2</version>
+            </plugin>
+        </plugins>
+    </reporting>
```
undx, line 212: Revert whitespace changes / maybe run build will fix indentation!

undx (approved):
LGTM (just see my comment on plugin pom)
Anyway, if possible because it may require some time, I suggest to add an UT/IT for testing the mojo:
- simulate a cloud-component-docker like module w/ a tesitng pom
- use a car with limited deps (ie not so big)
- run mojo
- check repo created by mojo (present deps, digest, etc.)

talend-component-maven-plugin/pom.xml
```diff
@@ -181,22 +185,27 @@
...
-                <pomInclude>*/pom.xml</pomInclude>
-              </pomIncludes>
-              <goals>
-                <goal>clean</goal>
-                <goal>verify</goal>
-              </goals>
-            </configuration>
           </execution>
         </executions>
+        <configuration>
```
quilicicf, line 190: The configuration block was not at the right location, it was moot. 
Learned it the hard way banging my head on the wall trying to understand why the script variables didn't show up in the post build script.

This probably means that the whole configuration was useless before (only default conf).

talend-component-maven-plugin/src/it/dependency/pom.xml
```diff
@@ -84,17 +82,32 @@
...
               <goal>dependencies</goal>
             </goals>
+            <phase>generate-sources</phase>
+          </execution>
+          <execution>
+            <id>prepare-repository</id>
+            <goals>
+              <goal>prepare-repository</goal>
+            </goals>
+            <phase>test</phase>
```
quilicicf, line 107: Not sure adding a new phase here is the right way but adding a new pom might be worse... 
I'll let you decide.

talend-component-maven-plugin/src/it/dependency/postbuild.groovy
```diff
@@ -0,0 +1,51 @@
+/**
```
quilicicf, line 1: The class `ValidateDependencies` seems not to be run.
I could not find anything in the invoker plugin's documentation hinting at Java classes from the IT projects being run.
I guess they should since the class is compiled and the `default-test` phase is run but I updated the class to throw and write crap to the FS to check whether it was run and it seems not.

I didn't fix it since this PR is reaaaaaaaaally dragging already but I'm almost positive that it's not run and thus useless.

quilicicf:
Added a test, that was a bumpy ride :weary:

talend-component-maven-plugin/src/it/dependency/pom.xml
undx: Fine for me

talend-component-maven-plugin/src/it/dependency/postbuild.groovy
undx: The test is not run because the pom lacks the failsafe plugin execution and test file is not following IT naming.
Anyway, not sure that this test is really helpful: Had a quick look at it not sure that it passes ...

talend-component-maven-plugin/pom.xml
undx: Good catch!

undx (approved):
great job, LGTM!

PR #558, opened 2022-03-30
fix(TCOMP-2102): Fix javadoc to repare build

quilicicf:
### Requirements

### Why this PR is needed?

The build on `master` is broken.

Fix the build.

### :warning: This fix is incomplete

We should always build the javadoc (or never, but here I believe we need it).

Currently, it is only built on 4 modules (added manually in their poms) when not on `master`: 

* `component-api`
* `component-runtime-junit`
* `component-runtime-beam-junit`
* `component-runtime-http-junit`

The issue is that I haven't found why yet. 

The javadoc is built:

* [partially on dev branches](https://jenkins-connectors.datapwn.com/job/component-runtime/job/build/job/master__TDI-2102_fixCloudComponentsBuild/7/consoleFull)
* [fully on master](https://jenkins-connectors.datapwn.com/job/component-runtime/job/build/job/master/522/consoleFull)

But the Maven commend that's run is the same in both cases: `mvn clean install -Dgpg.skip=true -Denforcer.skip=true -s .jenkins/settings.xml`

Can we dig a bit to make sure other people don't get caught in this trap? @undx


undx (approved):
LGTM

PR #560, opened 2022-04-04
fix(TCOMP-2112): make AvroRecordBuilderFactoryProvider not dependant to jackson-mapper-asl

undx:
https://jira.talendforge.org/browse/TCOMP-2112

- instead of testing a class present in classloader from jackson-mapper-asl, we test an internal class to beam module to be sure that the module is really in classloader and not eventually provided by a far transitive dependency as occurred in described jira issue.

### DIFF
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
@@ -15,8 +15,6 @@
  */
 package org.talend.sdk.component.runtime.beam.spi;
 
-import static java.util.Optional.ofNullable;
-
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 
@@ -46,11 +44,19 @@ public class AvroRecordBuilderFactoryProvider implements RecordBuilderFactoryPro
             return new AvroRecordBuilderFactory(containerId);
         default:
             try {
-                ofNullable(Thread.currentThread().getContextClassLoader())
-                        .orElseGet(ClassLoader::getSystemClassLoader)
-                        .loadClass("org.talend.sdk.component.runtime.beam.TalendIO");
+                // This part is tricky, we need to ensure that we've everything needed to instantiate an ARBF to user.
+                // Testing w/ loadClass won't be accurate enough as we may load all needed classes but w/ distinct cl
+                // like ConfigurableClassLoader vs Launcher$AppClassLoader.
+                // Safest way is to try a direct instantiation.
+                final Record r = new AvroRecordBuilderFactory(containerId)
+                        .newRecordBuilder()
+                        .withString("test", "instantiation")
+                        .build();
+                log.warn("[apply] AvroRecord instantiated.");
+                //
                 return new AvroRecordBuilderFactory(containerId);
-            } catch (final ClassNotFoundException | NoClassDefFoundError cnfe) {
+            } catch (final NoClassDefFoundError cnfe) {
+                log.warn("[apply] Error: ", cnfe);
                 log.info("component-runtime-beam is not available, skipping AvroRecordBuilderFactory ({})",
                         getClass().getName());
                 return new RecordBuilderFactoryImpl(containerId);
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
@@ -43,24 +43,31 @@ public class AvroRecordBuilderFactoryProvider implements RecordBuilderFactoryPro
         case "default":
             return new RecordBuilderFactoryImpl(containerId);
         case "avro":
+            if (!hasAvroRecordBuilderFactory()) {
+                log.warn(
+                        "AvroRecordBuilderFactoryProvider if forced by System property but seems not available, this may lead to issues.");
+            }
             return new AvroRecordBuilderFactory(containerId);
         default:
-            try {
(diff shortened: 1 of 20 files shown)


component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/record/RecordBuilderFactoryProviderTest.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.function.Function;
+
+import org.junit.jupiter.api.Test;
+import org.talend.sdk.component.api.service.record.RecordBuilderFactory;
+import org.talend.sdk.component.runtime.manager.ComponentManager;
+import org.talend.sdk.component.runtime.record.RecordBuilderFactoryImpl;
+
+public class RecordBuilderFactoryProviderTest {
```
clesaec, line 27: Should be better to enrich real test for [AvroRecordBuilderFactoryProvider](https://github.com/Talend/component-runtime/blob/master/component-runtime-beam/src/test/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProviderTest.java#L30) ? here, just testing a fake class.

To be complete, code 
```java
ofNullable(Thread.currentThread().getContextClassLoader())
                        .orElseGet(ClassLoader::getSystemClassLoader)
                        .loadClass("org.talend.sdk.component.runtime.beam.TalendIO");
```

can be put in a protected method :
```java
protected isTalendPipelineDesigner() {
    try {
        ofNullable(Thread.currentThread().getContextClassLoader())
                        .orElseGet(ClassLoader::getSystemClassLoader)
                        .loadClass("org.talend.sdk.component.runtime.beam.TalendIO");
        return true;
    } catch (final ClassNotFoundException | NoClassDefFoundError cnfe) {
        log
            .info(component-runtime-beam is not available, skipping AvroRecordBuilderFactory ({})",
              getClass().getName());
            return false;
    }
}
```
That make code clearer and easier to test

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
```diff
@@ -48,12 +48,11 @@ public RecordBuilderFactory apply(final String containerId) {
             try {
                 ofNullable(Thread.currentThread().getContextClassLoader())
                         .orElseGet(ClassLoader::getSystemClassLoader)
-                        .loadClass("org.codehaus.jackson.node.TextNode");
+                        .loadClass("org.talend.sdk.component.runtime.beam.TalendIO");
```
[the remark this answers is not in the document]
ozhelezniak-talend: My comment is more about: "Why do we have jackson-mapper-asl" as a test and not compile dependency if component-runtime-beam uses it.

clesaec (approved):
Nice, 
See comments : code can be made clearer and unit test more accuracy.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
```diff
@@ -46,14 +46,19 @@ public RecordBuilderFactory apply(final String containerId) {
             return new AvroRecordBuilderFactory(containerId);
         default:
             try {
-                ofNullable(Thread.currentThread().getContextClassLoader())
-                        .orElseGet(ClassLoader::getSystemClassLoader)
-                        .loadClass("org.codehaus.jackson.node.TextNode");
-                return new AvroRecordBuilderFactory(containerId);
+                final ClassLoader cl = ofNullable(Thread.currentThread().getContextClassLoader())
```
clesaec, line 49: While you have to rebase, extract the method that chose implementation :
```java
    @Override
    public RecordBuilderFactory apply(final String containerId) {
        switch (System.getProperty("talend.component.beam.record.factory.impl", "auto")) {
        case "memory":
        case "default":
            return new RecordBuilderFactoryImpl(containerId);
        case "avro":
            return new AvroRecordBuilderFactory(containerId);
        default:
            if (this.isAvroImpl()) {
                 return new AvroRecordBuilderFactory(containerId);
            }
            else {
                return new RecordBuilderFactoryImpl(containerId);
            }
        }
    }

protected boolean isAvroImpl() {
    try {
            final ClassLoader cl = ofNullable(Thread.currentThread().getContextClassLoader())
                        .orElseGet(ClassLoader::getSystemClassLoader);
            final Class<?> c1 = cl.loadClass("org.codehaus.jackson.node.TextNode");
            final Class<?> c2 = cl.loadClass("org.talend.sdk.component.runtime.beam.spi.record.AvroSchema");
            return c1.getClassLoader().equals(c2.getClassLoader());
     } catch (final ClassNotFoundException | NoClassDefFoundError cnfe) {
            log.info("component-runtime-beam is not available, skipping AvroRecordBuilderFactory ({}).",
                        getClass().getName());
                return false;
     }
}
```

clesaec (changes requested):
Change asked in same time of rebase

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/record/RecordBuilderFactoryProviderTest.java
undx: outdated

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/AvroRecordBuilderFactoryProvider.java
undx: Done +1

clesaec (approved):
:+1:

PR #562, opened 2022-04-07
fix(TDI-47694): [JDBC TCK]: Inject runtime context for tck connector runtime

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TDI-47694

https://jira.talendforge.org/browse/TDI-47694


wwang-talend:
in common javajet, we need do syn like this for create_connection/close part:
```
syn(lock) {
  1. inject context to service for connection/close
  2. call connection create or close by service
}
```

no need that syn for input or mapper/processor/standalone runner as they are not singleton

undx (commented):
Some small comments. 
And some questions
AfterVariables support:
 Isn't that a double usage use-case?  Can we have an unified solution?
 (maybe parametermodel maybe an issue...)

Studio injection:
 Could you define how will the _globalMap_ will be injected in Studio? With javajet I imagine...

Documentation:
 We need to refine documentation on this feature... Here's the place:
https://github.com/Talend/component-runtime/blob/master/documentation/src/main/antora/modules/ROOT/pages/javajet-to-componentkit.adoc

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInject.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Inject runtime context object to runtime input/processor/standalone runner object
+     */
+    public static void injectRuntimeContext(final Lifecycle lifecycle, final Map<String, Object> runtimeContext) {
+        if (lifecycle instanceof Delegated) {
+            final Object delegate = ((Delegated) lifecycle).getDelegate();
+
+            Class<?> currentClass = delegate.getClass();
+            while (currentClass != null && currentClass != Object.class) {
```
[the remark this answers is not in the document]
wwang-talend: sometimes, the input/processor class have parent class which may define the @RuntimeContext.

wwang-talend:
@undx 
https://github.com/Talend/component-runtime/pull/562/files#diff-e8dd30ce21874d45dadb5d3c7957bdbefff72ca815cf483bebdc37762204dd67
This is the whole code which we will also apply to common javajet, and have shown inject action which we will do in that junit.

undx (dismissed):
Nice! ;-)

undx:
Rebase needed

ozhelezniak-talend (changes requested):
injectLifecycle doesn't work as it was expected by its writer.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInject.java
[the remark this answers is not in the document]
undx: Did not catch it but Oleksandr is right, you don't use the correct variable.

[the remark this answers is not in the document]
wwang-talend: thanks @ozhelezniak-talend, fixed

wwang-talend:
Hi @ozhelezniak-talend , could you show more about:
https://github.com/Talend/component-runtime/pull/562#pullrequestreview-939348941
misunderstand

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/ReflectionUtils.java
```diff
@@ -0,0 +1,90 @@
...
+                            }
+                        }));
+    }
+
+    static Stream<Field> findFields(final Object delegate, final Class<? extends Annotation> marker) {
+        return findFields(delegate.getClass(), marker);
+    }
+
+    static Stream<Field> findFields(final Class delegateClass, final Class<? extends Annotation> marker) {
+        return Stream.of(delegateClass.getDeclaredFields()).filter(m -> m.isAnnotationPresent(marker)).peek(m -> {
```
ozhelezniak-talend, line 61: findFields(Object delegate, Class marker, ClassLoader loader) and this one have almost the same body. 
Is It expected that we have this code duplication and that we invoke this code not inside "callInLoader"?

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInjector.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Inject runtime context object to runtime input/processor/standalone runner object
+     */
+    public static void injectLifecycle(final Lifecycle lifecycle, final Map<String, Object> runtimeContext) {
+        if (lifecycle instanceof Delegated) {
+            final Object delegate = ((Delegated) lifecycle).getDelegate();
+
+            Class<?> currentClass = delegate.getClass();
+            while (currentClass != null && currentClass != Object.class) {
+                ReflectionUtils.findFields(currentClass, RuntimeContext.class).forEach(f -> {
```
ozhelezniak-talend, line 42: Yes, now we are handling all classes in the hierarchy of the instance: "delegate". Before that, we on each iteration over and over get the same exact class of the delegate object and its parent classes weren't searched for the fields.

ozhelezniak-talend: Please, look at the comment in the ReflectionUtils.findFields(currentClasss, RuntimeContext.class) method. 
Maybe we need to call its logic inside "callInLoader" too.

ozhelezniak-talend (commented):
It's ok, I just have a small warning about the additional method findFields that has Class in the list of arguments.

ozhelezniak-talend (approved):
looks lgtm according to previous commens

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/ReflectionUtils.java
undx: I do agree with Oleksandr. Why not let this method be a wrapper of `findFields(Object delegate, Class marker, ClassLoader loader)`?

undx: Also, the pattern of `findMethods` is identical... Maybe a generic method will avoid code duplication?

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInjector.java
```diff
@@ -0,0 +1,83 @@
...
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class RuntimeContextInjector {
+
+    /**
+     * Inject runtime context object to runtime input/processor/standalone runner object
+     */
```
undx, line 35: javadoc comment, please fill in correct content.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInjector.java
```diff
@@ -0,0 +1,83 @@
...
+        } else {
+            log.warn("Not supported implementation of lifecycle");
+        }
+    }
+
+    /**
+     * Inject runtime context object to runtime connection/close service/action object,
+     * maybe also commit/rollback if they are auto generated too,
+     * service is flat, no extends as common, so not do getSuperclass
+     */
```
undx, line 62: javadoc comment, please fill in correct content.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/RuntimeContextInjector.java
```diff
@@ -0,0 +1,83 @@
...
+                        return;
+                    } catch (final IllegalAccessException e) {
+                        throw new IllegalStateException(e);
+                    }
+                });
+
+                currentClass = currentClass.getSuperclass();
+            }
+        } else {
+            log.warn("Not supported implementation of lifecycle");
```
undx, line 54: May be throw an `IllegalArgumentException` here or change message to
`Unsupported Lifecycle implementation by {}` or smtg like that. 
(Even if Lifecycle interface contract does not declare anything about Delegated but each impl handles it...)

If exception option, remove any useless logging code, annotation and import.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/studio/AfterVariableExtracter.java
```diff
@@ -53,12 +43,11 @@ public static Map<String, Object> extractAfterVariables(final Lifecycle lifecycl
         if (lifecycle instanceof Delegated) {
             final Object delegate = ((Delegated) lifecycle).getDelegate();
 
-            ClassLoader classloader = Optional
-                    .ofNullable(Instance.get().find(lifecycle.plugin()).classloader())
-                    .orElseGet(() -> Thread.currentThread().getContextClassLoader());
-            return findMethods(delegate, AfterVariableContainer.class, classloader)
+            ClassLoader classloader = ReflectionUtils.getClassLoader(lifecycle);
```
undx, line 46: `final`

undx (approved):
lgtm

PR #563, opened 2022-04-14
chore(TCOMP-2121): change Record.withSchema

ozhelezniak-talend:
* an entry in a new schema can be wider than in old one

### Requirements
https://jira.talendforge.org/browse/TCOMP-2121


ozhelezniak-talend:
No need right now.

PR #564, opened 2022-04-14
fix(TCOMP-2111): support beam native mapper in ProducerFinder

undx:
https://jira.talendforge.org/browse/TCOMP-2121

- based on Chris' UT


component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -0,0 +1,161 @@
...
+    public Iterator<Record> find(final String familyName, final String inputName, final int version,
+            final Map<String, String> configuration) {
+        final ComponentInstantiator instantiator = getInstantiator(familyName, inputName);
+        final Mapper mapper = findMapper(instantiator, version, configuration);
+        try {
+            final Input input = mapper.create();
+            return iterator(input);
+        } catch (Exception e) {
+            log.warn("Component Kit Mapper instantiation failed, trying to wrap native beam mapper...");
+            final Object delegate = Delegated.class.cast(mapper).getDelegate();
```
clesaec, line 61: Is it Mapper or Input which contains delegated PTransform ?
(And exception occurs in iterator(input) call, so; all change could have been put here ?)
BTW, i can't find where TCK treat PTransform connectors.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/ProducerFinderImpl.java
```diff
@@ -76,7 +91,7 @@ static class InputIterator implements Iterator<Object> {
 
         private Object nextObject;
 
-        private boolean init = false;
+        private boolean init;
```
clesaec, line 94: Why remove explicit initialize .

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/ProducerFinderImpl.java
```diff
@@ -88,26 +103,26 @@ private static Object findNext(final Input input) {
 
         @Override
         public boolean hasNext() {
-            synchronized (this.input) {
+            synchronized (input) {
```
clesaec, line 106: Why remove "this." ? it help to read code.

clesaec (commented):
Well done,
Only some comments / questions ;

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/ProducerFinderImpl.java
undx: matter of taste... default value for `boolean` primitive is `false`.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/service/ProducerFinderImpl.java
undx: matter of taste also. As there's no ambiguity on the variable vs a member.

clesaec (approved):
On implicit vs explicit, my taste is to prefer explicit (init boolean explicitly, always use this. ...), i find it easier to read.
But, it's OK

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
undx: That's the mapper.
Don't get your final comment...

clesaec: ok, i see, name of S3Input class on cloud-component get me wrong, it's indeed a mapper
```java
@PartitionMapper(name = "S3Input")
@Documentation("This component reads data from S3.")
public class S3Input extends PTransform<PBegin, PCollection<Record>> {
```

PR #566, opened 2022-04-21
TCOMP-2123 : fix move after last

clesaec:
### Requirements
https://jira.talendforge.org/browse/TCOMP-2123

### DIFF
--- a/component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordBuilderImplTest.java
+++ b/component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordBuilderImplTest.java
@@ -31,6 +31,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordBuilderImplTest.java
```diff
@@ -56,6 +57,11 @@ void providedSchemaGetSchema() {
                         .build())
                 .build();
         assertEquals(schema, new RecordImpl.BuilderImpl(schema).withString("name", "ok").build().getSchema());
+
+        Schema.EntriesOrder e = new Schema.EntriesOrder(new RecordImpl.BuilderImpl().getCurrentEntries()
+                .stream()
+                .map(Schema.Entry::getName)
+                .collect(Collectors.toList()));
```
undx, line 64: did not catch, but what's the purpose?

clesaec: it appears after rebase from master, and with compile error; i just fix the error without asking purpose. I think it can be deleted.

PR #568, opened 2022-05-09
fix(TCOMP-2133): use baseVersion for snapshot artifacts

undx:
https://jira.talendforge.org/browse/TCOMP-2133

### DIFF
--- a/talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/ComponentDependenciesBase.java
+++ b/talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/ComponentDependenciesBase.java
@@ -143,7 +143,7 @@ public abstract class ComponentDependenciesBase extends AudienceAwareMojo {
     }
 
     protected String computeCoordinates(final Artifact artifact) {
-        return artifact.getGroupId() + ":" + artifact.getArtifactId() + ":" + artifact.getVersion();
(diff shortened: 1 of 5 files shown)


ozhelezniak-talend (approved):
lgtm

talend-component-maven-plugin/src/main/java/org/talend/sdk/component/maven/ComponentDependenciesBase.java
```diff
@@ -143,7 +143,7 @@ protected Artifact resolveArtifact(final Artifact artifact, final String classif
     }
 
     protected String computeCoordinates(final Artifact artifact) {
-        return artifact.getGroupId() + ":" + artifact.getArtifactId() + ":" + artifact.getVersion();
+        return artifact.getGroupId() + ":" + artifact.getArtifactId() + ":" + getVersion(artifact);
```
quilicicf, line 146: Doesn't that remove the `-SNAPSHOT` for snapshots? :thinking: 
Is that OK? 
I'm not sure about the impact of this one :sweat_smile:

undx: No, that's the opposite it's doesn't take timestamp but x.y.z-SNAPSHOT. Have a look at issue description.

quilicicf (approved):

PR #574, opened 2022-05-26
fix(TUP-35585): Zero install fail to install custom tck component in 801.

cmeng-talend:
https://jira.talendforge.org/browse/TUP-35585

### Requirements

### Why this PR is needed?
In https://jira.talendforge.org/browse/TUP-35585
There is a bug related to m2 repository location calculation in CarMain when studio/ci try to install the CAR, the CarMain can't calculate the correct m2 repository location; We tried to only fix the issue on studio side, to let the CarMain works correctly, but failed, so we have to update the codes in CarMain; Though we fix the issue in CarMain, but CARs which were created using old CarMain and deployed to prod env by users, studio/ci still can't run it correctly; to make studio/ci be compatible with CARs of different version, we need to know the CarMain version, so that we can use another logic in studio to install CAR correctly if the CarMain has the bug, and keep to run the CarMain of fixed version.

Adding new class **Versions**, so that we can know the component-tools version during runtime from codes, then we can add car bundler version info into the metadata file of Car, later we can know the car bundler and CarMain version from the metadata file inside the CAR.


kjwang-talend (approved):

cmeng-talend:
Done in TCOMP-2146

PR #578, opened 2022-06-10
fix(TCOMP-2166): enrich empty enum values

ozhelezniak-talend:
* empty string value converted to null enum value

### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2166


ozhelezniak-talend:
This problem helps us to find that some parts pass the empty value instead of skipping the hidden forms elements.
Looks like we don't need this fix.

PR #580, opened 2022-06-20
fix(TCOMP-2173): resolve maven settings

ozhelezniak-talend:
* use path separator for m2 paths

### Requirements
https://jira.talendforge.org/browse/TCOMP-2173


undx (approved):

ozhelezniak-talend:
> As remarked by @fdesiles , we may miss an UT for testing the modified method.

Didn't have an idea how to test it, will think about it.
The existed junit anyway returns false positive for this case. It uses fallback and it was ok with it. But with a custom config it fails. (Maven repo is not in the user home)

PR #581, opened 2022-06-20
chore(TDI-47867) Increase CI build visibility

fdesiles:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TDI-47867

Updated Jenkinsfile to:

1. Only send slack messages to #component-ci for Master and Maintenance builds.
2. Ping with "here" when a build is failing and the previous one was OK.


Jenkinsfile
```diff
@@ -27,6 +27,7 @@ final def isStdBranch = (env.BRANCH_NAME == "master" || env.BRANCH_NAME.startsWi
 final def tsbiImage = "artifactory.datapwn.com/tlnd-docker-dev/talend/common/tsbi/jdk11-svc-builder:2.9.27-20220331162145"
 final def jdk17Image= "artifactory.datapwn.com/tlnd-docker-dev/talend/common/tsbi/jdk17-svc-builder:2.9.27-20220331162145"
 final def podLabel = "component-runtime-${UUID.randomUUID().toString()}".take(53)
+final boolean isOnMasterOrMaintenanceBranch = env.BRANCH_NAME == 'master' || env.BRANCH_NAME.startsWith('maintenance/')
```
undx, line 30: use the `isStdBranch` variable, no need to create a new one.

undx (commented):
small comment.

Jenkinsfile
fdesiles: Good catch, missed it :)

PR #595, opened 2022-07-19
chore(TCOMP-2187): integer implicit constraint

ozhelezniak-talend:
* allow overwriting implicit annotation constraint

https://jira.talendforge.org/browse/TCOMP-2187


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/reflect/ParameterModelService.java
```diff
@@ -227,9 +232,18 @@ private Map<String, String> buildExtensions(final String name, final Type generi
...
+        final Set<? extends Class<? extends Annotation>> forbidImplicit = Stream.of(annotations)
+                .map(Annotation::annotationType)
+                .filter(uniqueAnnotations::contains)
+                .collect(toSet());
         return Stream
                 .concat(getReflectionAnnotations(type, annotations),
-                        implicitAnnotationsMapping.getOrDefault(type, emptyList()).stream());
+                        implicitAnnotationsMapping.getOrDefault(type, emptyList())
+                                .stream()
+                                .filter(it -> !forbidImplicit.contains(it.annotationType())));
```
olhava, line 246: Could it be optimized to use uniqueAnnotations instead of a separate forbidImplicit:
.filter(it -> !uniqueAnnotations.contains(it.annotationType())))

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/ParameterModelServiceTest.java
```diff
@@ -111,6 +112,54 @@ void charAsString() throws NoSuchMethodException {
         }
     }
 
+    @Test
+    void intAsLongByDefault() throws NoSuchMethodException {
```
olhava, line 116: Why the name of the method is '..AsLong', seems like you are testing int vs Integer constraints here

ozhelezniak-talend: yes, you are right. Originally it was another idea for this test. I'll rename it. Thank you.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/reflect/ParameterModelService.java
ozhelezniak-talend: It won't be the same. The variable name I may rename.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/reflect/ParameterModelService.java
```diff
@@ -227,9 +232,18 @@ private Map<String, String> buildExtensions(final String name, final Type generi
     }
 
     private Stream<Annotation> getAnnotations(final Type type, final Annotation[] annotations) {
+        // we have a few implicit constraints. those constraints can be overwritten by explicit one.
+        // to avoid ambiguous exception we should filter out implicit annotation constraints
+        final Set<Class<? extends Annotation>> uniqueAnnotations = new HashSet<>(Arrays.asList(Min.class, Max.class));
```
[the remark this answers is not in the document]
ozhelezniak-talend: I'll rename variable.

undx (approved):
Asked a question on non primitive case and validation on slack, if not the case lgtm

ozhelezniak-talend:
I added an additional test to cover the execution of implicit integer validation.
"Null" case is covered. It's ok. The null value is not the value for the current implementation of min and max validations.

undx (approved):
lgtm

PR #603, opened 2022-08-10
feat(TCOMP-2107): Implement a stop strategy for streaming input connectors

undx:
https://jira.talendforge.org/browse/TCOMP-2107

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
@@ -88,16 +88,16 @@ public class Streaming {
         final LocalConfiguration configuration = ofNullable(ContainerFinder.Instance.get().find(plugin))
                 .map(it -> it.findService(LocalConfiguration.class))
                 .orElseGet(defaultLocalConfiguration);
-        final Long maxReadRecords = ofNullable(configuration.get("talend.input.streaming.maxRecords"))
-                .map(Long::parseLong)
-                .orElse(ofNullable(System.getProperty(
-                        String.format("%s.talend.input.streaming.maxRecords", plugin), "-1"))
+        final Long maxReadRecords =
+                ofNullable(System.getProperty(String.format("%s.talend.input.streaming.maxRecords", plugin)))
+                        .map(Long::parseLong)
+                        .orElse(ofNullable(configuration.get("talend.input.streaming.maxRecords"))
                                 .map(Long::parseLong)
                                 .orElse(null));
-        Long maxActiveTime = ofNullable(configuration.get("talend.input.streaming.maxDurationMs"))
-                .map(Long::parseLong)
-                .orElse(ofNullable(System.getProperty(
-                        String.format("%s.talend.input.streaming.maxDurationMs", plugin), "-1"))
+        Long maxActiveTime =
+                ofNullable(System.getProperty(String.format("%s.talend.input.streaming.maxDurationMs", plugin)))
+                        .map(Long::parseLong)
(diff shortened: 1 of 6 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
```diff
@@ -0,0 +1,221 @@
...
+                            .map(Long::parseLong)
+                            .orElse(500L));
+        }
+    }
+
+    public static StopStrategy loadStopStrategy(final String plugin) {
+        final LocalConfiguration configuration = ofNullable(ContainerFinder.Instance.get().find(plugin))
+                .map(it -> it.findService(LocalConfiguration.class))
+                .orElseGet(defaultLocalConfiguration);
+        final Long maxReadRecords = ofNullable(configuration.get("talend.input.streaming.maxRecords"))
```
ozhelezniak-talend, line 91: Can I configure those configuration options in the component UI?
talend.input.streaming.maxRecords
talend.input.streaming.maxDurationMs

ozhelezniak-talend (commented):
We've lost an ability to endlessly run the component. If the max record and max timeout isn't set the component won't run at all.
Is it expected?

Do we guarantee that those two new options can't be equal -1?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
```diff
@@ -0,0 +1,221 @@
...
+
+        public StopConfiguration(final Long maxRecords, final Long maxTime, final Long start) {
+            maxReadRecords = maxRecords == null ? -1L : maxRecords;
+            maxActiveTime = maxTime == null ? -1L : maxTime;
+            started = start == null ? System.currentTimeMillis() : start;
+        }
+
+        @Override
+        public boolean isActive() {
+            return (maxReadRecords > -1) || (maxActiveTime > -1);
```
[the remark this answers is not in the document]
undx: By default, the stop stragegy conditions are both set to -1, this make the stragegy inactive. When it's inactive, the job runs endlessly. Only the retry stragegy can stop it!

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
undx: By default, first element checked is localconfiguration then system property. 
We can have a mapping with localconf on UI like this 
```
@Option
@DefaultValue("local_configuration:myfamily.model.key")
private String value;
```
As stated by documentation, some action need to be done to propagate information.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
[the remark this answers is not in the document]
ozhelezniak-talend: yes, you totally right. I misread the condition where the isActive is used.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
ozhelezniak-talend: yes, in this way we can have this value in UI. 
But as a user, I want to pass the value from the UI to the configuration to stop watch.

ozhelezniak-talend (approved):
LGTM
Great feature!

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
undx: https://jira.talendforge.org/browse/TCOMP-2226

PR #604, opened 2022-08-11
Use https instead of ssh

acatoire:
### Why this PR is needed?
To stop to use the old ssh credentials on the Jenkins instance


fdesiles (approved):
LGTM

PR #613, opened 2022-08-31
Feat TCOMP-2238	Upgrade jib-core to 0.16.0

yyin-talend:
https://jira.talendforge.org/browser/TCOMP-2238
### Requirements


undx (approved):
LGTM

PR #614, opened 2022-08-31
feat(TCOMP-2234): override blocking read process in streaming connectors

undx:
https://jira.talendforge.org/browse/TCOMP-2234

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
@@ -176,6 +176,12 @@ public class Streaming {
 
         boolean shouldStop(long read);
 
+        long getMaxReadRecords();
+
+        long getMaxActiveTime();
+
+        long getStarted();
+
     }
 
     @Data
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
@@ -172,15 +172,41 @@ public class Streaming {
 
     public interface StopStrategy {
 
+        /**
+         * Check if the stop strategy is active according specified conditions.
+         * 
+         * @return true if strategy is active, false otherwise.
+         */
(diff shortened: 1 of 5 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
```diff
@@ -176,6 +176,12 @@ public interface StopStrategy {
 
         boolean shouldStop(long read);
 
+        long getMaxReadRecords();
+
+        long getMaxActiveTime();
+
+        long getStarted();
```
ozhelezniak-talend, line 183: getStartedTime() Because I couldn't get what getStarted is. I even thing that we should change to isStarted but then I saw that it returns long. And I kept looking

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/StreamingInputImpl.java
```diff
@@ -65,21 +70,42 @@ protected Object readNext() {
...
-                return null;
-            }
             final RetryStrategy strategy = retryConfiguration.getStrategy();
             int retries = retryConfiguration.getMaxRetries();
             while (running.get() && retries > 0) {
-                final Object next = super.readNext();
+                Object next = null;
+                if (stopStrategy.isActive() && stopStrategy.getMaxActiveTime() > -1) {
+                    final ExecutorService executor = Executors.newSingleThreadExecutor();
+                    final Future<Object> reader = executor.submit(() -> super.readNext());
```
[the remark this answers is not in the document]
undx: Sure.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
undx: yep, that's a bit misleading! In fact, it means _lifecycle started at time n (system millis)_... Will refactor to get a better understanding like `getStartedAtTime`.

ozhelezniak-talend (approved):
LGTM

PR #616, opened 2022-08-31
chore(TCOMP-2252): call fallback when need

ozhelezniak-talend:
### Requirements
https://jira.talendforge.org/browse/TCOMP-2252

### Why this PR is needed?
I had a couple of rough moments when debugging the maven repository resolve and can't understand why the fallback was called when I saw that discover options are good.


undx (approved):
lgtm

PR #622, opened 2022-09-09
feat(TCOMP-2258): @Documentation to tooltips in uiSchema

undx:
https://jira.talendforge.org/browse/TCOMP-2258

### Requirements


component-server-parent/component-server/pom.xml
```diff
@@ -186,6 +186,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.talend.sdk.component</groupId>
+      <artifactId>component-form-core</artifactId>
+      <version>1.49.0-SNAPSHOT</version>
```
ozhelezniak-talend, line 192: project version?
instead of hardcoded

ozhelezniak-talend (approved):
lgtm. minor comment

component-server-parent/component-server/pom.xml
undx: arg, yep!

ozhelezniak-talend (approved):
lgtm

PR #629, opened 2022-09-16
Fix[TCOMP-2256]: Cannot invoke "javax.script.ScriptEngine.createBindings()" with Java 17

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2256

### Requirements


undx (commented):
If we use rhino instead of nashorn, we should remove the nashorn's references in project.
Please also, add an unit test like testing the script engine (w/ regexp for instance).

undx (approved):
LGTM

PR #637, opened 2022-09-29
chore(TCOMP-2269): Junit publisher fix 

acatoire:
https://jira.talendforge.org/browse/TCOMP-2269
### Why this PR is needed?
The Jenkins job missed JUnit result publishing.

Add JUnit publisher in the "always" step.


undx (approved):
lgtm

PR #646, opened 2022-10-21
Yyin/tcomp 2260 validations on nested obj

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2260

### Requirements


undx (approved):
LGTM

PR #648, opened 2022-10-25
fix(TCOMP-2277): Complete the TCK schema's entry properties

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2277

https://jira.talendforge.org/browse/TCOMP-2277


component-api/src/main/java/org/talend/sdk/component/api/record/EntryProperty.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.talend.sdk.component.api.record;
+
+public interface EntryProperty {
```
undx, line 18: In framework, I prefer `SchemaProperty` as naming for this class as we may also add some props at schema  level and not just enties.

undx (commented):
I think that we also need to unite keys in component-runtime-di...

component-api/src/main/java/org/talend/sdk/component/api/record/EntryProperty.java
wwang-talend: Hi @undx 
about "unite keys in component-runtime-di", that may need migration, will do it in TDI-48755 after TDI-48634 pass Pipeline team's usage.

wwang-talend: Hi @undx 
have them all, but i see we public something in tck document : "talend.studio.type", we need do migration on it? if that, advice to keep the old value for STUDIO_TYPE.

undx: `talend.studio.type` is distinct from `field.type`, it represents a studio type not a db type (ie meaning `id_String`, `id_Long`, etc.). But need to be taken in account anyway.

undx (dismissed):

PR #651, opened 2022-11-01
feat(TCOMP-2291): pass timeout & max records option

ozhelezniak-talend:
### Requirements
https://jira.talendforge.org/browse/TCOMP-2291

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2291

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
@@ -34,6 +34,8 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class Streaming {
 
+    public static final int MAX_DURATION_TIME_MS_GRACE_PERIOD = 3_000;
+
     private static Supplier<LocalConfiguration> defaultLocalConfiguration = () -> new LocalConfiguration() {
 
         @Override
--- a/component-tools/src/main/java/org/talend/sdk/component/tools/validator/OptionParameterValidator.java
+++ b/component-tools/src/main/java/org/talend/sdk/component/tools/validator/OptionParameterValidator.java
@@ -17,8 +17,11 @@ package org.talend.sdk.component.tools.validator;
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -35,11 +38,13 @@ import org.talend.sdk.component.api.input.PartitionMapper;
  */
 public class OptionParameterValidator implements Validator {
 
-    private static final Set<String> ALLOWED_OPTION_PARAMETERS = new HashSet<>();
+    private static final Map<String, Set<Class<?>>> ALLOWED_OPTION_PARAMETERS = new HashMap<>();
 
     static {
-        ALLOWED_OPTION_PARAMETERS.add(Option.MAX_DURATION_PARAMETER);
-        ALLOWED_OPTION_PARAMETERS.add(Option.MAX_RECORDS_PARAMETER);
+        ALLOWED_OPTION_PARAMETERS.put(Option.MAX_DURATION_PARAMETER,
+                new HashSet<>(Arrays.asList(int.class, Integer.class, long.class, Long.class)));
+        ALLOWED_OPTION_PARAMETERS.put(Option.MAX_RECORDS_PARAMETER,
+                new HashSet<>(Arrays.asList(int.class, Integer.class, long.class, Long.class)));
     }
 
     @Override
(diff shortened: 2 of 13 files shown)


undx (commented):
I think it may be wiser to give connector those informations before the connector has started working...
Some connectors may initialize first batch of readings before calling `@Producer /next()` method or may need those informations to elaborate a strategy knowning it may be stop (ie not  the case if un-stoppable ie records=-1 && duration=-1).
Thinking that giving those informations after the connector created, so in lifecyle using the @PostContruct method (if exists).
In that way that the connector knows all parameters and how to act according its tech specs.
Of course of all this should be available only if the stoppable flag is set.

Note sure about the grace period need in this case?

we will also need: 
- validation on mvn plugin 
- documentation (w/ warnings related to stop a streaming job...)

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/Streaming.java
```diff
@@ -34,6 +34,8 @@
 @Slf4j
 public class Streaming {
 
+    public static final int MAX_ACTIVE_TIME_GRACE_PERIOD = 3_000;
```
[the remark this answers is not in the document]
ozhelezniak-talend: It's just an almost random number. Almost it's like "42".
Less was sometimes not enough, and component-runtime interrupts before the JMS-broker library code can do smth.
Not sure what is the best way how to handle it. 
Balazs suggested some grace period before trying to interrupt. I kind of like it. 
Another idea was is to don't stop by component-runtime if we pass the parameter inside the connector's code.
But it might be a hidden problem when the connector's developer forgets to implement the logic on their side. 
Dunno. I 100% opened for suggestions here

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/input/InputImpl.java
```diff
@@ -79,15 +83,22 @@ public Object getDelegate() {
     }
 
     protected Object readNext() {
-        return doInvoke(this.next);
+        return doInvoke(this.next, nextParameters);
     }
 
     protected void init() {
         next = findMethods(Producer.class).findFirst().get();
```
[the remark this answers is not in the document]
ozhelezniak-talend: Okay

ozhelezniak-talend:
> Of course of all this should be available only if the stoppable flag is set.

I'm missing the flag inside StreamingInputImpl that can tell that this PartitionMapper isn't stoppable. 
I think if the configuration is NOT  isActive then all values will be **-1**  anyway.

pom.xml
```diff
@@ -746,6 +747,12 @@
       <version>${junit5.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <version>${mockito4.version}</version>
+      <scope>test</scope>
+    </dependency>
```
undx, line 755: Move this dep to DepMgnt and add it as dep in component-tools instead.

documentation/src/main/antora/modules/ROOT/pages/component-implementing-streaming.adoc
```diff
@@ -26,6 +26,27 @@ This default configuration implements a constant pause duration of 500 ms betwee
...
+@Version
+@PartitionMapper(name = "Input", infinite = true, stoppable = true)
+public class YourPartitionMapper {
+
+    @PostConstruct
+    public void init(@Option(Option.MAX_DURATION_PARAMETER) long maxDurationMs, @Option(Option.MAX_RECORDS_PARAMETER) long maxRecords ) {
+    	// connector's specific logic here
+    }
+    // other component's code
+}
```
undx, line 46: I think that it's more related to `@Emitter`/`@Producer` one not the one that impl `@PartitionMapper`. The producer needs to be informed not the mapper.

undx (commented):
end of day (and week) review, need to review more aware ;-) 
some comments.

pom.xml
ozhelezniak-talend: Just did it as junit5 was added.
I can add junit5 into depMgnt too.

documentation/src/main/antora/modules/ROOT/pages/component-implementing-streaming.adoc
ozhelezniak-talend: yes, you are totally right. 
end of the week documentation...

ozhelezniak-talend: fixed

pom.xml
ozhelezniak-talend: changed

component-tools/src/main/java/org/talend/sdk/component/tools/validator/OptionParameterValidator.java
```diff
@@ -0,0 +1,76 @@
...
+                .flatMap(m -> Stream.concat(
+                        Arrays.stream(m.getParameters())
+                                .filter(p -> !p.isAnnotationPresent(Option.class))
+                                .map(p -> "Parameter '" + p.getName()
+                                        + "' should be either annotated with @Option or removed"),
+                        Arrays.stream(m.getParameters())
+                                .filter(p -> p.isAnnotationPresent(Option.class))
+                                .filter(p -> !ALLOWED_OPTION_PARAMETERS.contains(p.getAnnotation(Option.class).value()))
+                                .map(p -> "Option value on the parameter '" + p.getName() + "' is not acceptable. "
+                                        + "Acceptable values: " + acceptableOptionValues())))
```
undx, line 67: maybe check also parameter's type if matching attented.

undx (approved):
LGTM, Great work!
Still have a small restriction on grace period time but if approved by Balázs...

ozhelezniak-talend:
> OptionParameterValidator.java

It's either a grace period or we should not stop Input components flow and let (believe) the component stop it.
I think we may change our behavior in the future. 
So we should put this warning in the documentation then. 
What do you think?

P.S. I'll add a type check in OptionValidation.

component-tools/src/main/java/org/talend/sdk/component/tools/validator/OptionParameterValidator.java
ozhelezniak-talend: Sure. will do

ozhelezniak-talend:
Hello @undx
Could you communicate your restrictions about the grace period in the ticket
https://jira.talendforge.org/browse/TCOMP-2291
We may remove it before the merge, or improve the approach.
I think just removing won't be a good idea because Component-runtime code will be in superposition and connector won't be able to respect timeout because component-runtime will interrupt it before.

undx (approved):

PR #652, opened 2022-11-03
fix(TCOMP-2292): Studio side, TCK Record/Schema builder should support Object type value like JDBC ResultSet object

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2292

https://jira.talendforge.org/browse/TCOMP-2292


wwang-talend:
TODO : add junit to DiRecordVisitorTest and DiRowStructVisitorTest

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitor.java
```diff
@@ -435,6 +435,8 @@ private void onArray(final Entry entry, final Collection value) {
     }
 
     private void onObject(final String name, final Object value) {
+        //in studio di, the object value is suitable to json string always? this code for construct the input tck record to tck processor, maybe is ok
+        //will adjust it if meet real requirement, then not toJson, only call "with(name, value)" api;
```
wwang-talend, line 439: should call with(schema.getEntry("name"), value) api

component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/RecordBuilderImplTest.java
```diff
@@ -251,6 +251,23 @@ void decimal() {
...
+                .withEntry(new SchemaImpl.EntryImpl.BuilderImpl()
+                        .withName("object")
+                        .withNullable(true)
+                        .withType(Type.STRING)
+                        .withProp("talend.studio.type", "id_Object")
+                        .build())
+                .build();
+        final RecordImpl.BuilderImpl builder = new RecordImpl.BuilderImpl(schema);
+        Object value = new Object();
+        final Record record = builder.with(schema.getEntry("object"), value).build();
```
undx, line 267: Try with some real world class...

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitor.java
undx: jsonb may help us a lot for ser/deser. See my comment on slack.

wwang-talend:
jsonb can't ser for jdbc resultset which have status and jsonb depend on get/is method, and some get/is method throw exception if status not match

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitor.java
```diff
@@ -435,6 +435,9 @@ private void onArray(final Entry entry, final Collection value) {
     }
 
     private void onObject(final String name, final Object value) {
+        // in studio di, the object value is suitable to json string always? this code for construct the input tck
+        // record to tck processor, maybe is ok
+        // will adjust it if not meet real requirement
         recordBuilder.withString(name, jsonb.toJson(value));
```
undx, line 441: may not fit against what is done in the PR...

wwang-talend: need to fix here for jdbc blob/clob type which map to id_Object, for example, suppose this all tck connector:
tmysqlinput(blob column)==>toracleoutput(blob column)
we can't call json.tojson here.

undx (approved):
LGTM

PR #653, opened 2022-11-04
feat(TCOMP-2285): allow jira ticket in artifact version

undx:
https://jira.talendforge.org/browse/TCOMP-2285

### DIFF
--- a/container/container-core/src/main/java/org/talend/sdk/component/container/ContainerManager.java
+++ b/container/container-core/src/main/java/org/talend/sdk/component/container/ContainerManager.java
@@ -51,6 +51,8 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.logging.Level;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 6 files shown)


container/container-core/src/main/java/org/talend/sdk/component/container/ContainerManager.java
```diff
@@ -268,6 +272,10 @@ public String buildAutoIdFromName(final String module) {
             if (autoId.isEmpty()) {
                 throw new IllegalArgumentException("Invalid name for plugin: " + module);
             }
+            final Matcher jiraTicket = versionWithJiraIssue.matcher(autoId);
```
ozhelezniak-talend, line 275: Those lines can be after. In case autoId is empty after the matcher check and substring
```
            if (autoId.isEmpty()) {
                throw new IllegalArgumentException("Invalid name for plugin: " + module);
            }
```

ozhelezniak-talend (commented):
It's cool to have this feature.
I left a comment, please, take a look.
Frankly, a bit confused with autoId.isEmpty check in the middle of the processing.

container/container-core/src/main/java/org/talend/sdk/component/container/ContainerManager.java
undx: sure, better place.

ozhelezniak-talend (approved):
LGTM

PR #661, opened 2022-11-28
Fix[TCOMP-2293]: Add Instant parameter type to withTimestamp method to avoid ms precision lose

yyin-talend:
[TCOMP-2293]  (https://jira.talendforge.org/browse/TCOMP-2293)
Add Instant parameter type to withTimestamp method to avoid ms precision lose

### Requirements


wwang-talend:
please check jenkins junits, guess the fix can't pass some test units.

TCK Record have two implements :
RecordImpl : which is basic one, major works for studio platform, which no need ser/desr most time, your fix is ok.
AvroRecord : which works for cloud platform with spark/hadoop/beam, which is supposed can works cross thread/vm or machine, so need to store data temp, need to ser/deser back, so seems your fix is not ok for that part, please make sure it.

component-runtime-beam/src/test/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecordTest.java
```diff
@@ -367,9 +367,11 @@ void pipelineDateTimeFields() throws Exception {
         final Date date = new Date(new java.text.SimpleDateFormat("yyyy-MM-dd").parse("2018-12-6").getTime());
         final Date datetime = new Date();
         final Date time = new Date(1000 * 60 * 60 * 15 + 1000 * 60 * 20 + 39000); // 15:20:39
+        final java.sql.Timestamp timestamp = java.sql.Timestamp.valueOf("2021-04-19 13:37:07.752345");
         builder.withDateTime("t_date", date);
         builder.withDateTime("t_datetime", datetime);
         builder.withDateTime("t_time", time);
+        builder.withTimestamp("t_timestamp", timestamp);
```
undx, line 374: value is not used/checked in this UT

undx (commented):
See my comment on used type.

As the record schema type is DATETIME, we should be able to call
- getDateTime("x") with an initial record made withInstant("x", v)
- getInstant("x")  with an initial record made withDateTime("x", zdt)

It will give a good UT to check the issue goal : do not loose nanosecs precision.
(check that in a case we loose precision, and in another one not)

AvroRecord seems lacking some implementation for this type and UT.
Please, also, increase UT : Record interface is used everywhere and should be bulletproof (trying so ;-) )

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroRecord.java
```diff
@@ -112,6 +115,10 @@ private Object directMapping(final Object value) {
         if (value instanceof Date) {
             return Date.class.cast(value).getTime();
         }
+        if (value instanceof Instant) {
+            //TCOMP-2293 support Instant with nano second
+            return new long[]{Instant.class.cast(value).getEpochSecond(), Instant.class.cast(value).getNano()};
```
wwang-talend, line 120: avro record works for cloud platform for ser/desr for spark/beam, and tck DATETIME will use this avro schema:
https://github.com/Talend/component-runtime/blob/117e2c64296c7fec809598606e8cb4ef784373ca/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchemaBuilder.java#L55
so guess this long[] not match that, and real can ser/deser for cloud platform?

undx: Why not extend with an union? Thus, we'll have union like {null, long, long[]}. WDYT @wwang-talend?

wwang-talend: Hi @undx , LGTM, no better idea now

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchemaBuilder.java
```diff
@@ -88,7 +88,13 @@ public class AvroSchemaBuilder implements Schema.Builder {
                             LogicalTypes
                                     .timestampMillis()
                                     .addToSchema(org.apache.avro.Schema.create(org.apache.avro.Schema.Type.LONG)),
-                            Schema.Type.DATETIME.name(), "true"))));
+                            Schema.Type.DATETIME.name(), "true"),
```
wwang-talend, line 91: no need to change for DATETIME_SCHEMA?

component-studio/component-runtime-di/src/test/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitorTest.java
```diff
@@ -68,6 +70,12 @@ void visit() {
                 .withString("date1", ZONED_DATE_TIME.toString())
                 .withDateTime("date2", ZONED_DATE_TIME)
                 .withLong("date3", ZONED_DATE_TIME.toInstant().toEpochMilli())
+                .withInstant(factory.newEntryBuilder()
+                        .withName("date4")
+                        .withType(Type.DATETIME)
+                        .withProp(STUDIO_TYPE, StudioTypes.DATE)
+                        .withProp(SchemaProperty.ORIGIN_TYPE, Instant.class.getName())
```
wwang-talend, line 77: SchemaProperty.ORIGIN_TYPE not for this

wwang-talend (changes requested):

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/MappingUtils.java
```diff
@@ -80,6 +83,20 @@ public static <T> Object coerce(final Class<T> expectedType, final Object value,
...
+            }
+            if (value instanceof long[]) {
+                final Instant instant = Instant.ofEpochSecond(((long[]) value)[0], ((long[]) value)[1]);
+                if (ZonedDateTime.class == expectedType) {
+                    return ZonedDateTime.ofInstant(instant, UTC);
+                }
+                if (Instant.class == expectedType) {
+                    return instant;
+                }
+            }
```
wwang-talend, line 98: add a expect type : java.util.Date

wwang-talend (approved):

wwang-talend (approved):

PR #678, opened 2022-12-29
fix(TCOMP-2330): A better way to check the JVM version

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2330

https://jira.talendforge.org/browse/TCOMP-2330

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/reflect/JavaVersion.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/reflect/JavaVersion.java
@@ -1,0 +1,129 @@
+/**
+ * Copyright (C) 2006-2022 Talend Inc. - www.talend.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/reflect/JavaVersion.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class JavaVersion {
+
+    /**
+     * Retrieve major version
+     *
+     * @param version
+     * @return major version
+     */
+    public static String getMajor(final String version) {
```
wwang-talend, line 26: advice make all methods to "private" if possible, and if for test, please remove the "public" to use default.

And LGTM, and not sure if we need so many methods here.

undx: Effectively, the need is only major but which can most can least ;-) May be useful in some cases some day...

undx (approved):

PR #679, opened 2022-12-30
fix(TCOMP-2165): [TCK]: tck connector need studio TIS license, as it hard depend on routines.system.Dynamic class

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2165

https://jira.talendforge.org/browse/TCOMP-2165


undx (dismissed):
LGTM

PR #680, opened 2023-01-06
fix(TCOMP-2338): do not leak talend.component.label prop in TCK Entry

anicolas:
See https://jira.talendforge.org/browse/TCOMP-2338

Let's discuss the possible implications of this change.

This is un-mergeable as it is, as it breaks other tests.

### DIFF
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
@@ -23,6 +23,7 @@ import static org.apache.avro.Schema.Type.UNION;
 import static org.talend.sdk.component.runtime.beam.avro.AvroSchemas.unwrapUnion;
 import static org.talend.sdk.component.runtime.record.SchemaImpl.ENTRIES_ORDER_PROP;
 
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/record/AvroSchema.java
```diff
@@ -184,7 +185,28 @@ private Entry fromAvro(final Field field) {
...
+     * <i>Note regarding the warning suppression: ideally, one should avoid using the {@link Field#getProps()}</i>
+     * method as it is deprecated...
+     */
+    @SuppressWarnings("deprecation")
     private static Entry buildFromAvro(final Field field, final Type type, final AvroSchema elementSchema) {
+        Map<String, String> tckEntryProps;
+
+        if (field.getProps().containsKey(KeysForAvroProperty.LABEL)) {
+            tckEntryProps = new HashMap<>(field.getProps());
+            tckEntryProps.remove(KeysForAvroProperty.LABEL);
```
undx, line 204: As discussed on slack, we shouldn't remove this property from others.

anicolas: I still think this is a bug.

However, I'll close the PR as we agreed we won't (can't?) fix it in a near future.

PR #681, opened 2023-01-11
feat(TCOMP-2336): Allow component-server to not fail on empty/not found component-registry descriptor

undx:
https://jira.talendforge.org/browse/TCOMP-2336


ozhelezniak-talend (approved):
lgtm

PR #682, opened 2023-01-11
fix(TCOMP-2339): remove static modifier

undx:
https://jira.talendforge.org/browse/TCOMP-2339


component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -171,21 +205,32 @@ private void sleep() {
...
+        private final int queueIndex;
+
+        public MyDoFn(final int queueIndex) {
+            this.queueIndex = queueIndex;
+        }
 
         @ProcessElement
-        public void processElement(final @Element Record record) {
-            boolean ok = QUEUE.offer(record);
+        public void processElement(ProcessContext context) {
```
undx, line 217: use `final` modifier.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -84,11 +88,21 @@ static class QueueInput extends LifecycleImpl implements Input, Iterator<Record>
...
         private Record next;
 
+        private final int queueIndex;
+
+        private Thread th;
+
         public QueueInput(final Object delegate, final String rootName, final String name, final String plugin,
-                final PTransform<PBegin, PCollection<Record>> transform) {
-            super(delegate, rootName, name, plugin);
+                          final PTransform<PBegin, PCollection<Record>> transform, int queueIndex) {
```
undx, line 96: use `final`.
```suggestion
                          final PTransform<PBegin, PCollection<Record>> transform, final int queueIndex) {
```

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -149,14 +171,26 @@ private PipelineResult runDataReadingPipeline() {
 
             try {
                 Thread.currentThread().setContextClassLoader(beamAwareClassLoader);
-
-                PipelineOptions options = PipelineOptionsFactory.create();
-                PushRecord pushRecord = new PushRecord();
+                DirectOptions options = PipelineOptionsFactory.as(DirectOptions.class);
+                options.setRunner(DirectRunner.class);
+                options.setTargetParallelism(10);
```
undx, line 176: Isn't that big? I don't know the implications on this parameter...

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -111,12 +125,20 @@ public Record next() {
...
-            Record record = QUEUE.poll();
-
+            System.out.println("Find Next, Thread " + Thread.currentThread().getId() +"; Queue size " + QUEUE.size());
+            System.out.flush();
+            Record record = QUEUE.get(this.queueIndex).poll();
+            int index = 0;
             while (record == null && (!end)) {
-                end = result.getState() != PipelineResult.State.RUNNING;
+                end = result != null && result.getState() != PipelineResult.State.RUNNING;
+                if (!end && index > 10) {
```
undx, line 134: I don't get the value _10_ here...

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -62,7 +65,8 @@ public Iterator<Record> find(final String familyName, final String inputName, fi
             log.warn("Component Kit Mapper instantiation failed, trying to wrap native beam mapper...");
             final Object delegate = Delegated.class.cast(mapper).getDelegate();
             if (PTransform.class.isInstance(delegate)) {
-                return new QueueInput(delegate, familyName, inputName, familyName, PTransform.class.cast(delegate));
+                QUEUE.add(new ArrayBlockingQueue<>(7, true));
```
undx, line 68: Why use _7_? (see me comment below on 10).
Maybe use the `CAPACITY` variable. See my comments on slack on performances.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -171,21 +205,32 @@ private void sleep() {
...
             while (!ok) {
-                if (QUEUE.size() >= CAPACITY) {
+                if (QUEUE.get(this.queueIndex).size() >= CAPACITY) {
+                    System.out.println("######### ERROR #######");
                     final String msg = String.format(
                             "Wrapper queue if full (capacity: %d). Consider increasing it according data with talend.beam.wrapper.capacity property.",
                             CAPACITY);
                     log.error("[processElement] {}", msg);
                     throw new IllegalStateException(msg);
                 }
```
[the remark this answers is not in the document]
undx: Maybe If using capacity  as commented above... but do we have to fail in that case?

undx (commented):
Some general comments:
- Please run checkstyle on module.
- I see that we do not remove the `static` modifier, is using an index such like `QUEUE.get(this.queueIndex)` fixes the issue ?
- module pom should be changed to allow `DirectRunner` to be in scope (scope in test). CI build fails.
- see comments on slack on performance.
Thanks for the work

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -15,44 +15,47 @@
...
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
 
 @Slf4j
 public class BeamProducerFinder extends ProducerFinderImpl {
 
     static final int CAPACITY = Integer.parseInt(System.getProperty("talend.beam.wrapper.capacity", "100000"));
 
-    static final Queue<Record> QUEUE = new ArrayBlockingQueue<>(CAPACITY, true);
+    private static List<Queue<Record>> QUEUE = new ArrayList<>();
```
clesaec, line 54: Use a Map instead a List to allow deletion of Queue after using it, as
`private static Map<UUID, Queue<Record>> QUEUEs = new ConcurrentHashMap<>();`
(then, a random generated UUID will replace the index)

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
clesaec: was one of my try, normally, it should not impact much performance; 
main thread consume the QUEUE while Beam Thread fill it.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
clesaec: here, main thread is waiting that beam thread put element in the QUEUE (with a sleep between each try).
After 10 try (can be parameterized, as the duration of sleep), we wait the full beam process to terminate.

component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
clesaec: May be too big indeed (should be less than QUEUE size), this is the number of sub-thread of Beam thread

yyin-talend (approved):

PR #688, opened 2023-02-07
feat(TCOMP-2369): make datetime picker configurable

undx:


constint (approved):
Tested OK for TDP (enables the needed fixes)

PR #689, opened 2023-02-09
fix(TCOMP-2371): [TCK JDBC]: Studio dynamic column metadata info : isKey should follow runtime dynamic object, not the input component's studio schema

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2371

https://jira.talendforge.org/browse/TCOMP-2371


undx (approved):
LGTM

PR #697, opened 2023-03-13
fix(TCOMP-1897): Bulk API - correct result type according API

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-1897

### Requirements

### DIFF
--- a/component-server-parent/component-server-model/src/main/java/org/talend/sdk/component/server/front/model/BulkResponses.java
+++ b/component-server-parent/component-server-model/src/main/java/org/talend/sdk/component/server/front/model/BulkResponses.java
@@ -39,5 +39,7 @@ public class BulkResponses {
         private Map<String, List<String>> headers;
 
         private byte[] response;
+
+        private String responseString;
     }
 }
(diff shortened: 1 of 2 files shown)


component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/front/BulkReadResourceImpl.java
```diff
@@ -195,6 +198,7 @@ private CompletableFuture<BulkResponses.Result> doExecute(final BulkRequests.Req
         final CompletableFuture<BulkResponses.Result> promise = new CompletableFuture<>();
         final InMemoryResponse response = new InMemoryResponse(() -> true, () -> {
             result.setResponse(outputStream.toByteArray());
+            result.setResponseString(new String(outputStream.toByteArray()));
```
undx, line 201: There are two responses in that case?

undx, line 201: Here we explicitly response with a string but expected type should be json as described in the API (https://github.com/Talend/component-runtime/blob/ee8fd4afd5a5dbb091b5b4dff900653f31a78d63/component-server-parent/component-server-api/src/main/java/org/talend/sdk/component/server/api/BulkReadResource.java#L37). We may perhaps have bad escaped value...

component-server-parent/component-server-model/src/main/java/org/talend/sdk/component/server/front/model/BulkResponses.java
```diff
@@ -39,5 +39,7 @@ public static class Result {
         private Map<String, List<String>> headers;
 
         private byte[] response;
+
+        private String responseString;
```
undx, line 43: This variable seems not needed, use response instead.

undx (commented):
Please, see comments and add tests for these changes.

undx (commented):
Please, also add some UT.

PR #700, opened 2023-03-20
fix(TCOMP-2404): Prevent API breakage on Record interface

undx:
https://jira.talendforge.org/browse/TCOMP2404


ozhelezniak-talend (approved):
lgtm

PR #715, opened 2023-04-13
feat(TCOMP-2409): add logicalType handling

undx:
POC for https://jira.talendforge.org/browse/TCOMP-2409


component-api/src/main/java/org/talend/sdk/component/api/record/SchemaProperty.java
```diff
@@ -33,4 +35,32 @@ public interface SchemaProperty {
...
+        }
+
+        public String key() {
+            return this.logicalType;
+        }
+
+        public Schema.Type storageType() {
+            return Schema.Type.DATETIME;
+        }
+    }
```
pierCo, line 64:
```suggestion
    enum LogicalType {

        DATE("date", Schema.Type.DATETIME),
        TIME("time", Schema.Type.DATETIME),
        TIMESTAMP("timestamp", Schema.Type.DATETIME),
        UUID("uuid", Schema.Type.STRING);

        private String logicalType;
        private Schema.Type type;

        LogicalType(final String logicalType, final Schema.Type type) {
            this.logicalType = logicalType;
            this.type = type;
        }

        public String key() {
            return this.logicalType;
        }

        public Schema.Type storageType() {
            return this.type;
        }
    }
```

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
```diff
@@ -514,6 +521,13 @@ public Builder withType(final Type type) {
                 return this;
             }
 
+            @Override
+            public Builder withLogicalType(final SchemaProperty.LogicalType logicalType) {
+                this.logicalType = logicalType;
+                this.props.put(SchemaProperty.LOGICAL_TYPE, logicalType.key());
+                return this;
+            }
```
pierCo, line 529: 🤔 why did you also put the `logicalType` into the `props` map ?

I ask that because it was not done for `type` or other attributes.

pierCo:
Any news about this feature ?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/SchemaImpl.java
undx: Because, the logicalType in the schema is held as a schema property, the logicalType type is only used in the builder.

PR #717, opened 2023-04-17
feat((TCOMP-2383): decrease log level when component not found

undx:
https://jira.talendforge.org/browse/TCOMP-2383


ypiel-talend (approved):
LGTM

PR #726, opened 2023-05-22
fix(TCOMP-2435): fix entrypoint.sh

undx:
https://jira.talendforge.org/browse/TCOMP-2435

### DIFF
--- a/component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/PropertiesSetup.java
+++ b/component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/PropertiesSetup.java
@@ -45,11 +45,12 @@ public class PropertiesSetup implements Meecrowave.ConfigurationCustomizer {
         System.setProperty("geronimo.opentracing.client.filter.request.skip", "true");
         System.setProperty("geronimo.opentracing.filter.skippedTracing.urls", ".*/login$,.*/decrypt/.*");
         System.setProperty("geronimo.opentracing.filter.skippedTracing.matcherType", "regex");
-        // environment dft ordinal 400
-        final String httpPort = System.getenv("TALEND_COMPONENT_SERVER_PORT");
-        if (httpPort != null) {
-            System.setProperty("http", httpPort);
-            configuration.setHttpPort(Integer.parseInt(httpPort));
+        // listening port order: `meecrowave cli --http=x` > `-Dhttp=x` > `-e TALEND_COMPONENT_SERVER_PORT=x`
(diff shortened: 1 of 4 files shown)


component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/PropertiesSetup.java
```diff
@@ -45,11 +45,12 @@ public void accept(final Configuration configuration) {
         System.setProperty("geronimo.opentracing.client.filter.request.skip", "true");
         System.setProperty("geronimo.opentracing.filter.skippedTracing.urls", ".*/login$,.*/decrypt/.*");
         System.setProperty("geronimo.opentracing.filter.skippedTracing.matcherType", "regex");
-        // environment dft ordinal 400
-        final String httpPort = System.getenv("TALEND_COMPONENT_SERVER_PORT");
-        if (httpPort != null) {
-            System.setProperty("http", httpPort);
-            configuration.setHttpPort(Integer.parseInt(httpPort));
+        // listening port order: `meecrowave cli --http=x` > `-Dhttp=x` > `-e TALEND_COMPONENT_SERVER_PORT=x`
+        final String port = System.getProperty("http") != null ? System.getProperty("http")
```
ozhelezniak-talend, line 49: Does "--http=x" is assigned to the system property "http" somewhere in the application?

images/component-server-image/src/main/jib/entrypoint.sh
```diff
@@ -0,0 +1,65 @@
...
+# Set Java classpath
+if [ -n "$CLASSPATH_OVERRIDE" ]; then
+    export CLASSPATH="$CLASSPATH_OVERRIDE"
+    unset CLASSPATH_OVERRIDE
+else
+    if [ -n "$CLASSPATH" ]; then
+        if [ -e "${TALEND_APP_HOME}/.tsbi.service.classpath" ]; then
+            export CLASSPATH="$CLASSPATH:$(cat ${TALEND_APP_HOME}/.tsbi.service.classpath)"
+        else
+            export CLASSPATH="$CLASSPATH"
```
ozhelezniak-talend, line 32: Why do we need this line?

component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/configuration/PropertiesSetup.java
undx: not in application directly. Usually from the launching environment (re2, ce,etc.)

PR #729, opened 2023-06-01
feat(TCOMP-2428): Stop deciphering vault secrets in migrations

undx:
https://jira.talendforge.org/browse/TCOMP-2428

### DIFF
--- a/component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ComponentResourceImplTest.java
+++ b/component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ComponentResourceImplTest.java
@@ -219,6 +219,32 @@ class ComponentResourceImplTest {
         assertEquals("true", migrated.get("migrated"));
     }
 
+    @Test
+    void migrateWithEncrypted() {
+        final Map<String, String> migrated = base
+                .path("component/migrate/{id}/{version}")
+                .resolveTemplate("id", client.getJdbcId())
+                .resolveTemplate("version", 1)
+                .request(APPLICATION_JSON_TYPE)
+                .header("x-talend-tenant-id", "test-tenant")
+                .post(entity(new HashMap<String, String>() {
+
+                    {
+                        put("configuration.url", "vault:v1:hcccVPODe9oZpcr/sKam8GUrbacji8VkuDRGfuDt7bg7VA==");
+                        put("configuration.username", "username0");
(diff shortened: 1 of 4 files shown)


component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ComponentResourceImplTest.java
```diff
@@ -219,31 +219,6 @@ void migrateFromStudioWs() {
         assertEquals("true", migrated.get("migrated"));
     }
 
-    @Test
-    void migrateWithEncrypted() {
```
ozhelezniak-talend, line 223: Why do we remove this test? I suppose this test still can validate that after the migration the vault value remained the same. Or do I misunderstood the idea of test?

component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ConfigurationTypeResourceImplTest.java
```diff
@@ -124,36 +121,6 @@ void migrateUnexpectedWithNPE() {
...
-                        put("configuration.connection.password",
-                                "vault:v1:hcccVPODe9oZpcr/sKam8GUrbacji8VkuDRGfuDt7bg7VA==");
-                    }
-                }, APPLICATION_JSON_TYPE))
-                .readEntity(Map.class);
-        assertEquals("true", config.get("configuration.migrated"));
-        assertEquals("5", config.get("configuration.size"));
-        assertEquals("vault:v1:hcccVPODe9oZpcr/sKam8GUrbacji8VkuDRGfuDt7bg7VA==", config.get("configuration.url"));
-        assertEquals("username0", config.get("configuration.username"));
-        assertEquals("test", config.get("configuration.connection.password"));
```
ozhelezniak-talend, line 154: Here we can check that instead of test we still have vault value, like for configuration.url. It's hard to get the test wirhout knowing structure before and after

component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ComponentResourceImplTest.java
undx: effectively, can keep it but adapted.

ozhelezniak-talend (approved):
lgtm

PR #740, opened 2023-06-19
fix(TCOMP-2468): look for a component across all plugins

ozhelezniak-talend:
* ComponentInstatiator looks for component name in all matched plugins not only in the first one

### Requirements


undx (approved):
lgtm

PR #753, opened 2023-07-07
doc(TCOMP-1897): correct bulk api doc

acatoire:
https://jira.talendforge.org/browse/TCOMP-1897
This commit goal is to correct the bulk doc typo.
Usage of MARKDOWN carry return (double spaces).
It also corrects an API test.


undx (approved):
lgtm

PR #758, opened 2023-07-25
fix(TCOMP-2505): mvn talend-component:web not works

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2505

https://jira.talendforge.org/browse/TCOMP-2505


undx (commented):
This PR does not match the issue mentioned.
`component-server-image` module is not related to maven plugin which provides the goal `web` described in the issue.

PR #779, opened 2023-08-14
fix(TCOMP-2520): Pass datepattern for Strings

dmytro-sylaiev:
### Requirements

### Why this PR is needed?
Bug: TCOMP-2520

Add date pattern for String columns in Dynamic (when Dynamic column was read from file)


component-studio/component-runtime-di/src/test/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitorTest.java
```diff
@@ -15,21 +15,6 @@
...
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.talend.sdk.component.api.record.SchemaProperty.IS_KEY;
-import static org.talend.sdk.component.api.record.SchemaProperty.SIZE;
-import static org.talend.sdk.component.api.record.SchemaProperty.PATTERN;
-import static org.talend.sdk.component.api.record.SchemaProperty.SCALE;
-import static org.talend.sdk.component.api.record.SchemaProperty.STUDIO_TYPE;
-
```
undx, line 29: revert import changes.

undx (approved):
lgtm

component-studio/component-runtime-di/src/test/java/org/talend/sdk/component/runtime/di/record/DiRowStructVisitorTest.java
undx: statics should be in first according component-runtime style rules based on those mentioned above.

PR #786, opened 2023-09-04
fix(TCOMP-2526):

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2526

CVE fix: com.google.guava:guava:jar:31.1-jre:runtime; upgrade to 32.1.2-jre

### Requirements


undx (approved):
lgtm

PR #789, opened 2023-09-14
feat(TCOMP-2459) Cover dynamic_values action

lxia-talend:
https://jira.talendforge.org/browse/TCOMP-2459
Cover dynamic_values action


acatoire (approved):
LGTM

PR #787, opened 2023-09-06
fix(TCOMP-2374): DIRecordVisitor to fill Studio Dynamic object's column name by tck Schema.Entry.getName which must follow tck name rule

wwang-talend:
### Requirements

* Any code change adding any logic **MUST** be tested through a unit test executed with the default build
* Any API addition **MUST** be done with a documentation update if relevant 

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2374

### What does this PR adds (design/code thoughts)?
https://jira.talendforge.org/browse/TCOMP-2374

PR #790, opened 2023-09-14
Revert "fix(TCOMP-2374): DIRecordVisitor to fill Studio Dynamic object's column name by tck Schema.Entry.getName which must follow tck name rule"

undx:
Reverts Talend/component-runtime#787


fdesiles (approved):
LGTM

PR #799, opened 2023-09-26
fix(COMP-2541): Improve performance on dynamic column

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2541

https://jira.talendforge.org/browse/TCOMP-2541

### DIFF
--- a/component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitor.java
+++ b/component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitor.java
@@ -32,16 +32,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.ZonedDateTime;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.OptionalDouble;
-import java.util.OptionalInt;
-import java.util.OptionalLong;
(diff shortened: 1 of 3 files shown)


component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitor.java
```diff
@@ -31,16 +32,7 @@
...
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.OptionalDouble;
-import java.util.OptionalInt;
-import java.util.OptionalLong;
-import java.util.Set;
+import java.util.*;
```
undx, line 35: do not use star import please.

undx (commented):
just one comment

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitor.java
undx, line 35: do not use star import please.

undx (commented):
just one comment. Otherwise LGTM

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/record/DiRecordVisitor.java
wwang-talend: thanks a lot, just fixed.

undx (approved):
LGTM

PR #801, opened 2023-09-26
chore(TDI-49155): Add Sonar to GitHub

acatoire:
### Why this PR is needed?
Correct the sonar ci integration
https://jira.talendforge.org/browse/TDI-49155

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,10 +28,11 @@ final def sonarCredentials = usernamePassword( credentialsId: 'sonar-credentials
 final def keyImportCredentials = usernamePassword(credentialsId: 'component-runtime-import-key-credentials', usernameVariable: 'KEY_USER', passwordVariable: 'KEY_PASS')
 final def gpgCredentials = usernamePassword(credentialsId: 'component-runtime-gpg-credentials', usernameVariable: 'GPG_KEYNAME', passwordVariable: 'GPG_PASSPHRASE')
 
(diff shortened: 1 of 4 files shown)


fdesiles (changes requested):

fdesiles (dismissed):
LGTM

Jenkinsfile
```diff
@@ -18,20 +18,49 @@ import java.time.LocalDateTime
...
+        credentialsId: 'component-runtime-gpg-credentials',
+        usernameVariable: 'GPG_KEYNAME',
+        passwordVariable: 'GPG_PASSPHRASE')
+
+// In PR environment, the branch name is not valid and should be swap with pr name.
+final String pull_request_id = env.CHANGE_ID
+final String branch_name = pull_request_id != null ? env.CHANGE_BRANCH : env.BRANCH_NAME
 
 // Job config
-final String slackChannel = 'components-ci'
```
fdesiles, line 32: slackChannel variable should not be suppressed.

fdesiles (changes requested):

Jenkinsfile
acatoire: It was suppressed because of copy past from connectors, we should use the env var now. Especially with future slack migration risk for name change.
I corrected all usage.

fdesiles (approved):
LGTM

PR #809, opened 2023-10-12
feat(TCOMP-2535): allow optional row

undx:
https://jira.talendforge.org/browse/TCOMP-2535


ozhelezniak-talend (approved):
lgtm
Do we plan to update documentation with this ew option?

PR #811, opened 2023-10-19
fix(TDI-46171): support custom connection/close component icon for tck connector in studio

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2552

https://jira.talendforge.org/browse/TCOMP-2552


wwang-talend:
sonar advice fix, and not care the duplicated string issues which exists before, need more careful refactor.

component-server-parent/component-server-api/src/main/java/org/talend/sdk/component/server/api/ComponentResource.java
```diff
@@ -135,6 +135,23 @@ Response icon(@PathParam("id") @Parameter(name = "id",
             description = "Component icon identifier.",
             in = PATH) String id);
 
+    @GET
+    @Path("searchIcon/{familyId}/{iconKey}")
```
undx, line 139: I tend to prefer not changing api prefix, rather use `icon`. smtg like `icon/custom/...`.
```suggestion
    @Path("icon/custom/{familyId}/{iconKey}")
```

wwang-talend: @undx fixed

undx (dismissed):
lgtm

undx (approved):

PR #814, opened 2023-10-24
feat(TCOMP-2554): Restore extraBuildParams in release script

acatoire:
Restore extraBuildParams in release script
https://jira.talendforge.org/browse/TCOMP-2554

### Why this PR is needed?
The CI was using env var to pass args to the release script.
After CI evolution this env var was not accessible anymore, this PR restore the functionality using script parameter instead of env var.

Add parameter to release.sh

### DIFF
--- a/.jenkins/scripts/release.sh
+++ b/.jenkins/scripts/release.sh
@@ -20,9 +20,13 @@ set -xe
 # Parameters:
 # $1: Branch
 # $2: current version
+# $3: extra build args for all mvn cmd
 main() {
   local branch="${1?Missing branch}"
   local currentVersion="${2?Missing project version}"
+  local extraBuildParams="${3}"
+
+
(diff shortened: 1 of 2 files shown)


.jenkins/scripts/release.sh
```diff
@@ -20,9 +20,13 @@ set -xe
 # Parameters:
 # $1: Branch
 # $2: current version
+# $3: extra build args for all mvn cmd
 main() {
   local branch="${1?Missing branch}"
   local currentVersion="${2?Missing project version}"
+  local extraBuildParams="${3}"
+
+
```
undx, line 29: why empty lines?

undx (approved):

.jenkins/scripts/release.sh
acatoire: It was to add space between parameters vars and others vars

PR #818, opened 2023-10-31
feat(TCOMP-2562): remove grpc from beam-runners-spark-3

undx:
https://jira.talendforge.org/browse/TCOMP-2562


ozhelezniak-talend (approved):
lgtm

PR #822, opened 2023-11-21
fix(TCOMP-2565): support origin type pass for studio dynamic type

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2565

https://jira.talendforge.org/browse/TCOMP-2565


undx (approved):
lgtm

acatoire:
The PR is adding a sonar code smell, is it possible to fix it before merge?

wwang-talend:
sonar think too many assert in one test unit method, it exists before already, here only add a new one. We can adjust it in a new workitem, here focus on current issue.

PR #823, opened 2023-11-22
fix(TCOMP-2573): null value configuration accepted

ozhelezniak-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2573


undx:
Fix will be done in a studio integration pull-request.

PR #827, opened 2023-11-30
feat(TCOMP-2578): remove geronimo-opentracing from component-server

undx:
https://jira.talendforge.org/browse/TCOMP-2578


ozhelezniak-talend (approved):
lgtm

PR #842, opened 2024-01-19
fix(TCOMP-2475): Migration from higher version not protected for a

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2475

connector

### Requirements


undx (commented):
LGTM
See with Axel for api-tester IT

acatoire:
IT test are now Green on this test, so the ticket is ready for me. I did the validation on the branch.

component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/front/ComponentResourceImplTest.java
```diff
@@ -179,6 +178,23 @@ void migrate() {
         assertEquals("true", migrated.get("migrated"));
     }
 
+    @Test
+    void no_migrate() {
```
undx, line 182: Please avoid underscore in method's name.
```suggestion
    void protectHigherMigration() {
```
or 
```suggestion
    void noMigration() {
```

undx:
> IT tests are now Green on this test, so the ticket is ready for me. I did the validation on the branch. Regarding the Unit test, we have two cases (if (version > and if version ==) to have a complete coverage of branches could you add the other unit test?

There's already an UT testing the case. But, yes we can add more tests on subject @yyin-talend.

undx (approved):
lgtm

PR #853, opened 2024-02-13
fix(TCOMP-2626): fix "null" literal in mapString

undx:
https://jira.talendforge.org/browse/TCOMP-2626

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/MappingUtils.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/MappingUtils.java
@@ -65,11 +65,6 @@ public class MappingUtils {
             }
         }
 
-        // we store decimal by string for AvroRecord case
-        if ((expectedType == BigDecimal.class) && String.class.isInstance(value)) {
-            return new BigDecimal(String.class.cast(value));
-        }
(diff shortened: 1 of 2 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/MappingUtils.java
```diff
@@ -65,11 +65,6 @@ public static <T> Object coerce(final Class<T> expectedType, final Object value,
             }
         }
 
-        // we store decimal by string for AvroRecord case
-        if ((expectedType == BigDecimal.class) && String.class.isInstance(value)) {
```
ozhelezniak-talend, line 69: Hm. If it was already supported by the line 212 in mapString, why did we have this if statement?

ozhelezniak-talend (approved):
lgtm

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/MappingUtils.java
undx: It passed the review. Reviewer did not catch it as mapString method wasn't modified I think (and it's maybe me :-D )

PR #859, opened 2024-02-23
fix(TCOMP-2631): use URLDecoder.decode method

undx:
https://jira.talendforge.org/browse/TCOMP-2631

### DIFF
--- a/component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
+++ b/component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
@@ -20,7 +20,6 @@ import static java.util.stream.Collectors.joining;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -32,6 +31,7 @@ import java.io.OutputStream;
 import java.io.Writer;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.time.LocalDateTime;
@@ -390,7 +390,7 @@ public class CarMain {
             if ("jar".equals(url.getProtocol())) {
                 final String spec = url.getFile();
                 final int separator = spec.indexOf('!');
-                return new File(decode(new URL(spec.substring(0, separator)).getFile()));
+                return new File(URLDecoder.decode(new URL(spec.substring(0, separator)).getFile(), "UTF-8"));
 
             } else if ("file".equals(url.getProtocol())) {
                 return toFile(resourceName, url);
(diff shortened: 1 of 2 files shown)


wwang-talend:
no test case to test it, advice to add if not complex to add case.

component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
```diff
@@ -390,8 +391,8 @@ private static File jarFromResource(final ClassLoader loader, final String resou
             if ("jar".equals(url.getProtocol())) {
                 final String spec = url.getFile();
                 final int separator = spec.indexOf('!');
-                return new File(decode(new URL(spec.substring(0, separator)).getFile()));
-
+                return new File(
+                        URLDecoder.decode(new URL(spec.substring(0, separator)).getFile(), "UTF-8"));
```
ozhelezniak-talend, line 395: We can use the constant here instead of string "UTF-8"
java.nio.charset.StandardCharsets#UTF_8
That method has an overload with param as Charset.
Same comment for all invocations of this method

component-tools/src/test/java/org/talend/sdk/component/tools/exec/CarMainTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+class CarMainTest {
+
+    @Test
+    void urlDecoderDecode() throws Exception {
+        final String expected = "/C:/Users/Cédric/Talend/Talend-Studio-V8.0.1/test-1.55.0.car";
+        final String car = new String("file:/C:/Users/C%c3%a9dric/Talend/Talend-Studio-V8.0.1/test-1.55.0.car"
+                .getBytes(StandardCharsets.ISO_8859_1));
+        final String path = new URL(car).getFile();
+        final File file = new File(URLDecoder.decode(path, "UTF-8"));
```
ozhelezniak-talend, line 35: For the test too
to use the Charset constant from JDK

ozhelezniak-talend (commented):
Could you use a constant instead of string literal?

component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
undx: Using StandardCharsets isn't java 8 compliant.

component-tools/src/test/java/org/talend/sdk/component/tools/exec/CarMainTest.java
undx: same.

component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
ozhelezniak-talend: Why? The class has existed since 1.7

ozhelezniak-talend (approved):
LGTM

component-tools/src/main/java/org/talend/sdk/component/tools/exec/CarMain.java
undx: the method `decode(String, Charset)` was introduced with jdk10.

PR #863, opened 2024-03-04
fix(TCOMP-2633): load nested package-info

ozhelezniak-talend:
### Requirements


undx (commented):
LGTM. If feasible, could you add an UT?

ozhelezniak-talend:
> LGTM. If feasible, could you add an UT?

Right now I don't see how

undx (approved):

PR #888, opened 2024-05-06
fix(TCOMP-2707): correct workflow and exception throwing

undx:
https://jira.talendforge.org/browse/TCOMP-2707


ozhelezniak-talend (approved):
LGTM...

PR #891, opened 2024-05-24
fix(TCOMP-2712):  fix create cache issue

undx:
https://jira.talendforge.org/browse/TCOMP-2712


component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/service/jcache/FrontCacheKeyGenerator.java
```diff
@@ -66,7 +65,7 @@ private Stream<Object> getContextualKeys() {
                             headers.getHeaderString(HttpHeaders.ACCEPT_ENCODING));
         } catch (Exception e) {
             log.debug("[getContextualKeys] context not applicable: {}", e.getMessage());
-            return Stream.of(UUID.randomUUID());
+            return Stream.empty();
```
ozhelezniak-talend, line 68: Hm, we changed the behavior. 
Not sure how it will influence the flow =)

ozhelezniak-talend (approved):
LGTM, just a small comment about behavior change, I hope it'll be fine, I'm not familiar with this part of functionality

component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/service/jcache/FrontCacheKeyGenerator.java
undx: In fact, this fixes a buggy idea for studio side... making never hitting the cache key!

PR #893, opened 2024-05-28
fix(TCOMP-2732): dynamic not overwritten

ozhelezniak-talend:
* prevent values for dynamic schema to be overwritten
* https://jira.talendforge.org/browse/TCOMP-2732

### Requirements


undx (approved):
LGTM
We did not used `routines.system.Dynamic` directly (thus wrapper) due to TOS limitations. But as TOS is a dead product  :-( , should have no issue

PR #896, opened 2024-06-03
feat(TCOMP-2741): revert remove static modifier in beam producer finder

undx:
https://jira.talendforge.org/browse/TCOMP-2741
Revert remove static modifier in BeamProducerFinder TCOMP-2339

### DIFF
--- a/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
+++ b/component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
@@ -20,14 +20,11 @@ import java.io.Serializable;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Queue;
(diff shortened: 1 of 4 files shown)


component-runtime-beam/src/main/java/org/talend/sdk/component/runtime/beam/spi/BeamProducerFinder.java
```diff
@@ -79,7 +72,7 @@ Object writeReplace() throws ObjectStreamException {
         return new SerializableService(plugin, ProducerFinder.class.getName());
     }
 
-    static class QueueInput implements Iterator<Record>, Serializable {
+    static class QueueInput extends LifecycleImpl implements Input, Iterator<Record> {
```
ozhelezniak-talend, line 76: not sure that this comment is applicable, but maybe Sonar find smth, that we didn't see

ozhelezniak-talend (approved):
LGTM (as revert)

PR #897, opened 2024-06-06
feat(TCOMP-2744): add @ReadOnly annotation

undx:
https://jira.talendforge.org/browse/TCOMP-2744


component-form/component-form-core/src/main/java/org/talend/sdk/component/form/internal/converter/impl/widget/AbstractWidgetConverter.java
```diff
@@ -172,6 +172,9 @@ protected UiSchema newOrphanSchema(final PropertyContext<?> ctx) {
         if (Boolean.parseBoolean(ctx.getProperty().getMetadata().getOrDefault("documentation::tooltip", "false"))) {
             schema.setTooltip(ctx.getProperty().getMetadata().get("documentation::value"));
         }
+        if (Boolean.parseBoolean(ctx.getProperty().getMetadata().getOrDefault("ui::readonly", "false"))) {
+            schema.setReadOnly(true);
```
ozhelezniak-talend, line 176: wow, we had this property???

ozhelezniak-talend (approved):
LGTM

component-form/component-form-core/src/main/java/org/talend/sdk/component/form/internal/converter/impl/widget/AbstractWidgetConverter.java
undx: It's in the UiSchema specification so already existed.

PR #898, opened 2024-06-06
feat(TCOMP-2475): Migration from higher version not protected for a connector

undx:
https://jira.talendforge.org/browse/TCOMP-2475


component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/front/ComponentResourceImpl.java
```diff
@@ -566,9 +566,15 @@ public Map<String, String> migrate(final String id, final int version, final Map
                         .status(Status.NOT_FOUND)
                         .entity(new ErrorPayload(COMPONENT_MISSING, "Didn't find component " + id))
                         .build()));
-        if (version > comp.getVersion() || version == comp.getVersion()) {
+        if (version > comp.getVersion()) {
```
ozhelezniak-talend, line 569: Why not >=?
If the config version == the version, we shouldn't migrate it, right?

ozhelezniak-talend: Why do we call migration if versions ==

ozhelezniak-talend (commented):
I have a question about the call of migration when the versions are equal.
And also I don't see a protection when the incoming version is higher than the one that is registered in component-runtime.
And I thought one of the points is to have some notification about it. Warning at least. 
Maybe I'm missing a point.

component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/front/ComponentResourceImpl.java
undx: There's a comment on jira issue concerning studio's migration handling.
Migration fails in that case and the job cannot be opened. 
Same behavior observed in slack topic before the revert.

undx:
> I have a question about the call of migration when the versions are equal. And also I don't see a protection when the incoming version is higher than the one that is registered in component-runtime. And I thought one of the points is to have some notification about it. Warning at least. Maybe I'm missing a point.

Answered to equals in comment.
There's a protection and warning preventing component migration call from higher version here
https://github.com/Talend/component-runtime/pull/898/files/4c91f86d3b6a5fdadda74bc4bb3a49450bbd7b98#diff-c897c2077dde44ed22410fa091a71724ed24211cfde6dd20348e11bf49a4d30aR570-R571

ozhelezniak-talend:
> > I have a question about the call of migration when the versions are equal. And also I don't see a protection when the incoming version is higher than the one that is registered in component-runtime. And I thought one of the points is to have some notification about it. Warning at least. Maybe I'm missing a point.
> 
> Answered to equals in comment. There's a protection and warning preventing component migration call from higher version here https://github.com/Talend/component-runtime/pull/898/files/4c91f86d3b6a5fdadda74bc4bb3a49450bbd7b98#diff-c897c2077dde44ed22410fa091a71724ed24211cfde6dd20348e11bf49a4d30aR570-R571

So it will a warning...
Oh, don't know about that
It's really looks like smth is fishy is going on when someone tries to run the configuration that is higher than is known (registered in) by component-runtime

component-server-parent/component-server/src/main/java/org/talend/sdk/component/server/front/ComponentResourceImpl.java
ozhelezniak-talend: Oh my...
Okay

undx:
> > > I have a question about the call of migration when the versions are equal. And also I don't see a protection when the incoming version is higher than the one that is registered in component-runtime. And I thought one of the points is to have some notification about it. Warning at least. Maybe I'm missing a point.
> > 
> > 
> > Answered to equals in comment. There's a protection and warning preventing component migration call from higher version here https://github.com/Talend/component-runtime/pull/898/files/4c91f86d3b6a5fdadda74bc4bb3a49450bbd7b98#diff-c897c2077dde44ed22410fa091a71724ed24211cfde6dd20348e11bf49a4d30aR570-R571
> 
> So it will a warning... Oh, don't know about that It's really looks like smth is fishy is going on when someone tries to run the configuration that is higher than is known (registered in) by component-runtime

Warning and call to migrate is skipped.

ozhelezniak-talend (approved):
LGTM

ozhelezniak-talend (approved):
lgtm

PR #910, opened 2024-07-11
Item(TCOMP-2766): check svg with log.error (#909)

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2766

### Requirements


acatoire (approved):
LGTM

PR #914, opened 2024-08-08
fix(TCOMP-2066): Job - afterGroup not called on Lifecycle::stop

undx:
https://jira.talendforge.org/browse/TCOMP-2066


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/AutoChunkProcessor.java
```diff
@@ -55,6 +58,7 @@ public void flush(final OutputFactory outs) {
 
     @Override
     public void stop() {
+        flush(remains);
```
ozhelezniak-talend, line 61: It looks like we are doing something wrong with this OutputFactory parameter being saved from the parameters.
Can we invoke flush before "stop()" on the client side?

component-runtime-manager/src/test/java/org/talend/test/FileOutput.java
```diff
@@ -42,18 +47,36 @@ public class FileOutput implements Serializable {
...
-    public void length(final JsonObject data) throws IOException {
+    private List<JsonObject> buffer;
+
+    @BeforeGroup
+    public void init() {
+        buffer = new ArrayList<>();
+    }
+
+    @AfterGroup
+    public void commit(final Collection<JsonObject> records) {
```
ozhelezniak-talend, line 58: Is it possible to have in the same Output a method with ElementListener and AfterGroup method with records as an input?

undx: Yes, you can manage it manually (elementlistener will receive record by record). Here, we have the shortess one.
```
    @ElementListener
    public void bufferize(final JsonObject object) {
        buffer.add(object);
    }

    @AfterGroup
    public void commit() {
        // saves buffered records at once (bulk)
    }
``` 
or just
 ```
    @AfterGroup
    public void commit(final Collection<Record> records) {
        // saves records
    }
```

ozhelezniak-talend: yes, but in your example you use both at the same time
my question is about using it at the same time

undx: oups, missed the signature of the commit method... right

ozhelezniak-talend (approved):
lgtm

PR #916, opened 2024-08-19
fix(TCOMP-1859):Field with name not avro compliant, fail if the schema is given to the RecordBuilder

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-1859

### Requirements


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -83,13 +83,14 @@ private RecordImpl(final Map<String, Object> values, final Schema schema) {
 
     @Override
     public <T> T get(final Class<T> expectedType, final String name) {
-        final Object value = values.get(name);
+        final String entryName = Schema.sanitizeConnectionName(name);
+        final Object value = values.get(entryName);
```
undx, line 87: I don't think that it's wise to apply a sanitzation again for getting values.
We cannot ensure that the result of ` Schema.sanitizeConnectionName(name)` will return the same sanitize made in builder...

What if I change a bit one of your test below like that?
```
       final Entry entry1 = new EntryImpl.BuilderImpl() //
                .withName("1 name a") //
                .withType(Type.DATETIME) //
                .withDefaultValue(null) //
                .build();
        final Entry entry2 = new EntryImpl.BuilderImpl() //
                .withName("2 name a") //
                .withType(Type.DOUBLE) //
                .withDefaultValue(11.11) //
                .build();
...
```
If 
- you call ` Schema.sanitizeConnectionName("1 name a")`  you'll get `_name_a`.
- you call ` Schema.sanitizeConnectionName("2 name a")`  you'll get `_name_a`.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -166,7 +167,7 @@ private BuilderImpl(final List<Schema.Entry> entries, final Map<String, Object>
 
         @Override
         public Object getValue(final String name) {
-            return this.values.get(name);
+            return this.values.get(Schema.sanitizeConnectionName(name));
```
undx, line 170: Certainly, same comment here...

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -200,10 +201,11 @@ public Builder with(final Entry entry, final Object value) {
 
         @Override
         public Entry getEntry(final String name) {
+            final String entryName = Schema.sanitizeConnectionName(name);
```
undx, line 204: Certainly, same comment here...

component-runtime-impl/src/test/java/org/talend/sdk/component/runtime/record/SchemaImplTest.java
```diff
@@ -182,6 +194,115 @@ void testRecordWithMetadataFields() {
...
+
+    @Test
+    void testNameSpace3() {
+        final Entry entry1 = new EntryImpl.BuilderImpl() //
+                .withName("1 name a") //
+                .withType(Type.DATETIME) //
+                .withDefaultValue(null) //
+                .build();
+        final Entry entry2 = new EntryImpl.BuilderImpl() //
+                .withName("2 name b") //
```
undx, line 278: changing with value "2 name b" to "2 name a" is correct but will lead to issues for example...

undx (commented):
See my comments.
I think again that sanitize naming is really a point which may lead to many issues!
If we really want using sanitize, we should provide some method for looking for entries with rawname...

PR #920, opened 2024-08-27
fix(TCOMP-2798): ConfigurableClassLoader is slow when load class from a signed jar

wwang-talend:
### Requirements

### Why this PR is needed?
https://jira.talendforge.org/browse/TCOMP-2798

provide a api to allow client user to use urlconnection cache in that classloader

### DIFF
--- a/component-api/src/main/java/org/talend/sdk/component/api/service/dependency/Resolver.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/service/dependency/Resolver.java
@@ -83,7 +83,7 @@ public interface Resolver {
          */
         Collection<String> resolvedDependencies();
 
-        default void setUseURLConnectionCaches(final boolean useCaches) {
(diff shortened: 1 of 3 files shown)


wwang-talend:
about junit, as a signed jar can reproduce it, and that jar not in .m2, also that is performance issue

component-api/src/main/java/org/talend/sdk/component/api/service/dependency/Resolver.java
```diff
@@ -82,5 +82,8 @@ interface ClassLoaderDescriptor extends AutoCloseable {
          * @return the dependencies who matched the resolution and were used to create the classloader.
          */
         Collection<String> resolvedDependencies();
+
+        default void setUseURLConnectionCaches(final boolean useURLConnectionCaches) {
```
wwang-talend, line 86: maybe no need to use this "default", as only ClassLoaderDescriptorImpl implement this interface?

PR #922, opened 2024-09-02
fix(TCOMP-2798): enable cacheable class loading

undx:
https://jira.talendforge.org/browse/TCOMP-2798


wwang-talend (dismissed):

container/container-core/src/main/java/org/talend/sdk/component/classloader/ConfigurableClassLoader.java
```diff
@@ -108,6 +108,10 @@ public class ConfigurableClassLoader extends URLClassLoader {
 
     private final URLClassLoader classLoaderFromClasspath;
 
+    @Getter
+    private final List<String> cacheableClasses =
```
wwang-talend, line 112: when this parameter not set, this will return a array with lenght 1 : [""], is not expected in my view.

wwang-talend (dismissed):

wwang-talend (approved):

PR #923, opened 2024-09-02
fix(TCOMP-2800): use calling module classloader as secondary loader

undx:
https://jira.talendforge.org/browse/TCOMP-2800

### DIFF
--- a/component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/InternalizationServiceTest.java
+++ b/component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/InternalizationServiceTest.java
@@ -17,8 +17,10 @@ package org.talend.sdk.component.runtime.manager.service;
 
 import static java.util.Collections.emptyMap;
 import static java.util.Optional.ofNullable;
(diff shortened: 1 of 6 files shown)


component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/service/InternalizationServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.talend.sdk.component.runtime.input.Mapper;
+import org.talend.sdk.component.runtime.manager.ComponentManager;
+import org.talend.sdk.component.runtime.manager.asm.PluginGenerator;
+
+public class InternalizationServiceTest {
+
+    private final PluginGenerator pluginGenerator = new PluginGenerator();
+
+    @Test
+    void wrongCL(@TempDir final File temporaryFolder) {
```
ozhelezniak-talend, line 36: Where is assert? 
What is testing here? 
If problem - an exception will be thrown, and if not - than it's ok, smth like that?

ozhelezniak-talend (approved):
lgtm

PR #929, opened 2024-09-30
fix(TCOMP-2845): add themed icons in starter generator

undx:
https://jira.talendforge.org/browse/TCOMP-2845


component-starter-server/src/main/java/org/talend/sdk/component/starter/server/service/facet/component/ComponentGenerator.java
```diff
@@ -254,9 +255,14 @@ private Stream<StringTuple3> toProperties(final String prefix, final Collection<
         });
     }
 
-    private FacetGenerator.InMemoryFile generateIcon(final String iconResourcesDirectory, final String iconName) {
```
ozhelezniak-talend, line 257: generateIcon(s)?

ozhelezniak-talend (dismissed):
lgtm

PR #930, opened 2024-09-30
fix(TCOMP-2846): fix JobImpl when no output factory available

undx:
https://jira.talendforge.org/browse/TCOMP-2846


ozhelezniak-talend (approved):
lgtm

PR #936, opened 2024-10-16
feat(QTDI-113): Support Java 17 compilation

undx:
https://qlik-dev.atlassian.net/browse/QTDI-113


pom.xml
```diff
@@ -150,6 +150,9 @@
   <properties>
     <scm.repository>talend/component-runtime</scm.repository>
 
+    <maven.compiler.source>17</maven.compiler.source>
```
ozhelezniak-talend, line 153: When do we plan to use this new component-runtime in Studio?
We can't use it before R2025-02, we should support Java 8 and Java11 before R2025-02

talend-component-maven-plugin/pom.xml
```diff
@@ -59,7 +60,7 @@
     <dependency>
       <groupId>org.apache.maven.plugin-tools</groupId>
       <artifactId>maven-plugin-annotations</artifactId>
-      <version>${mvn-minimal.version}</version>
+      <version>${maven-plugin-annotations.version}</version>
```
ozhelezniak-talend, line 63: Where do we use "mvn-minimal.version"?

ozhelezniak-talend (dismissed):
lgtm, but please see comments

talend-component-maven-plugin/pom.xml
undx: It's used in requireMavenVersion for enforcer plugin.

pom.xml
undx: Yeah, aware on the delays. We may postpone to next rel.

PR #938, opened 2024-10-18
feat(QTDI-739): remove bom

undx:
https://jira.talendforge.org/browse/TCOMP-2418
https://qlik-dev.atlassian.net/browse/QTDI-739


ozhelezniak-talend (dismissed):
lgtm
looks like the idea to use BOM, and it's hierarchy is official gone =(

ozhelezniak-talend (approved):

PR #942, opened 2024-10-28
fix(QTDI-749): remove png check

yyin-talend:
https://qlik-dev.atlassian.net/browse/QTDI-749

### Requirements


component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
```diff
@@ -183,25 +178,20 @@ private String validateIcon(final Icon annotation, final Collection<String> erro
                     prefixes.add(s + File.separator + ICONS + "dark" + File.separator + icon);
                 });
                 svgs = prefixes.stream().map(s -> new File(s + ".svg")).collect(toSet());
-                pngs = prefixes.stream().map(s -> new File(s + "_icon32.png")).collect(toSet());
             }
 
-            svgs.stream()
```
ozhelezniak-talend, line 189: this part was checking icons for legacy folder
and on Cloud we still used legacy icons, not dark and light theme
@undx Do you have some remarks about it?

component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
```diff
@@ -183,25 +178,20 @@ private String validateIcon(final Icon annotation, final Collection<String> erro
...
+            List<File> missingSvgs = svgs.stream().filter(f -> !f.exists()).collect(toList());
+            if (!missingSvgs.isEmpty()) {
+                errors.addAll(missingSvgs.stream()
                         .map(p -> String.format(
-                                "No icon: '%s' found, did you create - or generated with svg2png in resources?",
+                                "No '%s' found.",
                                 stripPath(p)))
                         .collect(toList()));
-                return "Missing icon(s) in resources.";
+                return "Missing icon(s) in resources:";
```
ozhelezniak-talend, line 194: why do we put ":" at the end?

component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
yyin-talend: "svg.streams()" is on line 187 now.

component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
```diff
@@ -183,25 +178,20 @@ private String validateIcon(final Icon annotation, final Collection<String> erro
...
             }
-            List<File> missingPngs = pngs.stream().filter(f -> !f.exists()).collect(toList());
-            if (!missingPngs.isEmpty()) {
-                errors.addAll(missingPngs.stream()
+
+            List<File> missingSvgs = svgs.stream().filter(f -> !f.exists()).collect(toList());
+            if (!missingSvgs.isEmpty()) {
+                errors.addAll(missingSvgs.stream()
                         .map(p -> String.format(
-                                "No icon: '%s' found, did you create - or generated with svg2png in resources?",
```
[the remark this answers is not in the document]
yyin-talend: This issue is just removing the check for missing png. No need to check png any more.

component-tools/src/main/java/org/talend/sdk/component/tools/ComponentValidator.java
yyin-talend: Because if this message shown, means that some icons were missing, next will list some missed icons, so here is ":" , means following list some icons.

undx (approved):

PR #945, opened 2024-11-07
fix(QTDI-763) issue to enable Tacokit

yyin-talend:
Parent - https://qlik-dev.atlassian.net/browse/QTDI-763
Task - https://qlik-dev.atlassian.net/browse/QTDI-835

### Requirements


undx (commented):
Could you add a test for that method?

undx (approved):
LGTM
In my opinion, in the UT, you should check returned value to be sure it used studio osgi location...

PR #948, opened 2024-11-15
fix(QTDI-747) Add junit coverage for svg2png

yyin-talend:
[QTDI-747](https://qlik-dev.atlassian.net/browse/QTDI-747)

### Requirements

[QTDI-747]: https://qlik-dev.atlassian.net/browse/QTDI-747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/component-tools/src/test/java/org/talend/sdk/component/tools/SVG2PngTest.java
+++ b/component-tools/src/test/java/org/talend/sdk/component/tools/SVG2PngTest.java
@@ -53,13 +53,18 @@ class SVG2PngTest {
     }
 
     @Test
-    void exception() {
-        String path = "D:\\works\\qlik";
+    void exceptionWithNotExistPath() {
+        String path = "not/existing/path";
         Assertions.assertThrows(java.lang.IllegalStateException.class, () -> {
             new SVG2Png(Paths.get(path), true, log).run();
         });
+    }
+
+    @Test
+    void exceptionWithWrongMethod() {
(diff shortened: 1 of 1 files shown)


component-tools/src/test/java/org/talend/sdk/component/tools/SVG2PngTest.java
```diff
@@ -49,4 +51,15 @@ void convert() throws IOException {
         assertTrue(Files.exists(darkExpected));
         assertTrue(Files.exists(lightExpected));
     }
+
+    @Test
+    void exception() {
```
acatoire, line 56: Maybe a more clear name, like `raise_exception_with_not_existing_path`

component-tools/src/test/java/org/talend/sdk/component/tools/SVG2PngTest.java
```diff
@@ -49,4 +51,15 @@ void convert() throws IOException {
...
+        String path = "D:\\works\\qlik";
```
acatoire, line 57: When I want to set an error path, I like it to be more “unreal”
```suggestion
        String path = "not/existing/path";
```

acatoire (changes requested):

acatoire (approved):
LGTM

PR #950, opened 2024-11-22
feat(QTDI-875): extend validation to object

undx:
https://qlik-dev.atlassian.net/browse/QTDI-875


ozhelezniak-talend (approved):
Kind of yes
But maybe we can check if the object is Datastore, Dataset, or object marked with an annotation that has type = "configuration"
it will be correct check.

PR #951, opened 2024-11-25
chore(QTDI-877): support conf/setting.xml in MavenDecrypter

ozhelezniak-talend:
https://qlik-dev.atlassian.net/browse/QTDI-877
### Requirements

### DIFF
--- a/component-runtime-testing/component-runtime-junit/src/main/java/org/talend/sdk/component/maven/MavenDecrypter.java
+++ b/component-runtime-testing/component-runtime-junit/src/main/java/org/talend/sdk/component/maven/MavenDecrypter.java
@@ -55,6 +55,8 @@ public class MavenDecrypter {
 
     private static final String M2_HOME = "M2_HOME";
 
+    private static final String MAVEN_HOME = "MAVEN_HOME";
+
     private static final String USER_HOME = "user.home";
 
     private static final String FILE_SETTINGS = "settings.xml";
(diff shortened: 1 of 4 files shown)


undx (approved):
lgtm.
Thinking also that getter should not be an issue.
cliir maven plugin should be able to be configured to skip this in module.

undx (dismissed):
See sonar comments. Mark'em as fixed for the relevants otherwise lgtm

component-runtime-testing/component-runtime-junit/src/main/java/org/talend/sdk/component/maven/MavenDecrypter.java
```diff
@@ -82,21 +100,63 @@ public Server find(final String serverId) {
...
-                .orElseGet(() -> new File(System.getProperty("user.home"), ".m2"));
+                .orElseGet(() -> new File(System.getProperty(USER_HOME), ".m2"));
+    }
+
+    private static List<File> findSettingsFiles() {
+        return Stream.of(
+                new File(getM2(), FILE_SETTINGS),
+                ofNullable(System.getenv(M2_HOME))
+                        .map(File::new)
+                        .map(it -> new File(it, "conf/" + FILE_SETTINGS))
```
undx, line 156: AFAIR, `conf/` is for maven home instead. Not sure it's in m2 home.

undx (dismissed):
one small comment to check otherwise lgtm

component-runtime-testing/component-runtime-junit/src/main/java/org/talend/sdk/component/maven/MavenDecrypter.java
undx: yep, confirmed https://maven.apache.org/settings.html

ozhelezniak-talend: yes, and M2_HOME it's a path to maven installation directory
https://www.baeldung.com/java-maven-environment-variables
for maven 1 it was MAVEN_HOME
for maven 2 they renamed in to M2_HOME
what is for maven 3 - I don't know

ozhelezniak-talend: Okay, I see in that explanation, that they start to call it again MAVEN_HOME
I can change it to MAVEN_HOME
or we can support both
What do you think?

undx: As we handle multiple locations, let's put maven home as latest one.
Few chances that user uses that one...

undx (approved):

PR #954, opened 2024-11-28
fix(QTDI-329) fix trivy

acatoire:
https://qlik-dev.atlassian.net/browse/QTDI-329
remove invalid line :  'apk add curl' because curl is on our image.
Move scripts to files. 

### Requirements


jzhao-talend (approved):
LGTM

PR #966, opened 2024-12-13
Yyin/QTDI-653 add LastGroup annotation to Processor's afterGroup method.

yyin-talend:
https://qlik-dev.atlassian.net/browse/QTDI-653

### Requirements


undx (commented):
Hello,
some comments below.

Why changing the sample-connector, it's not necessary. You can create UT elsewhere in project.
Concerning UT, I think that you should enforce tests on this feature.
- check backward compatibility
- check state when using lastgroup
- etc.

We also need to change validation for aftergroup. We should not be able to
handle both signatures and check for that.

I don't see why `component-tools/src/test/resources/org/talend/test/valid/httpclient/Messages.properties`
has some changes?

QueueOutput as some todo tasks that were used for debugging...

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/AutoChunkProcessor.java
```diff
@@ -26,4 +28,31 @@ public class AutoChunkProcessor extends org.talend.sdk.component.runtime.manager
     public AutoChunkProcessor(final int chunkSize, final Processor processor) {
```
undx, line 28: Please, create an UT to test theses changes. See UTs in this module.

undx (commented):
See comment.
Validation is missing in maven plugin. The ModelVisitor checks for runtime parameter not during build.

undx (approved):
lgtm

PR #969, opened 2024-12-17
feat(QTDI-693): manage nested jars

undx:
https://qlik-dev.atlassian.net/browse/QTDI-693

component-runtime:
- https://github.com/Talend/component-runtime/pull/969

master:
- https://github.com/Talend/tdi-studio-se/pull/10638
- https://github.com/Talend/tdi-studio-ee/pull/3346
- https://github.com/Talend/tcommon-studio-se/pull/8132
- https://github.com/Talend/tesb-studio-ee/pull/1440

maintenance/8.0.2 (used for building test studio) :
- https://github.com/Talend/tdi-studio-se/pull/10842
- https://github.com/Talend/tdi-studio-ee/pull/3446
- https://github.com/Talend/tcommon-studio-se/pull/7806
- https://github.com/Talend/tesb-studio-ee/pull/1423


container/container-core/src/main/java/org/talend/sdk/component/container/ContainerManager.java
```diff
@@ -111,13 +113,19 @@ public ContainerManager(final DependenciesResolutionConfiguration dependenciesRe
...
-                .filter(Files::exists)
-                .orElseGet(() -> PathFactory.get(System.getProperty("user.home", "")).resolve(".m2/repository"));
 
-        if (log.isDebugEnabled()) {
-            log.debug("Using root repository: " + this.rootRepositoryLocation.toAbsolutePath());
+        Path rootRepo = ofNullable(dependenciesResolutionConfiguration.getRootRepositoryLocation())
+                .filter(Files::exists)
+                .orElseGet(() -> PathFactory.get(System.getProperty("user.home")).resolve(".m2/repository"));
+        // if we've defaulted to user home m2 (fallback), we want to check if we're in running in a fatjar
+        if (PathFactory.get(System.getProperty("user.home")).resolve(".m2/repository").equals(rootRepo)) {
```
ozhelezniak-talend, line 121: M2 repository path can be ovewritten by <localRepository> in settings.xml file
=> ${user.home}/.m2/repository might not be a path to the repository.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -1115,16 +1117,41 @@ public void onCreate(final Container container) {
...
+                        while (urls.hasMoreElements()) {
+                            final URL url = urls.nextElement();
+                            // ensure we scan the correct classes of plugin in the nested repository
+                            if ("nested".equals(url.getProtocol())
+                                    && url.getPath().contains(container.getRootModule())) {
+                                try (final BufferedReader reader =
+                                        new BufferedReader(new InputStreamReader(url.openStream()))) {
+                                    config.load(reader);
+                                    filter = createScanningFilter(config);
+                                    alreadyScannedClasses = config.getProperty("classes.list");
```
ozhelezniak-talend, line 1134: should we "break" when we find the nested resource and create "filter" and "alreadyScannedClasses"?
because otherwise in case if we find more, the previous values will be overwritten
should we accumulate it maybe?

undx: We could add a break statement to stop iteration.
Anyway, normally we won't find any other resource as we check `url.getPath().contains(container.getRootModule())`

ozhelezniak-talend (approved):

PR #991, opened 2025-02-10
chore(QTDI-1189): bump component-runtime to 1.78.0-SNASPHOT

undx:
https://qlik-dev.atlassian.net/browse/QTDI-1189


ozhelezniak-talend (approved):

documentation/src/main/java/org/talend/runtime/documentation/Generator.java
```diff
@@ -185,7 +185,6 @@ public static void main(final String[] args) {
                 log.info("System is offline, skipping jira changelog and github contributor generation");
             } else {
                 tasks.register(() -> generatedContributors(generatedDir, args[5], args[6]));
-                tasks.register(() -> generatedJira(generatedDir, args[1], args[2], version));
```
fdesiles, line 188: Hello,
Why do we change documentation generator in a version bump ticket please?

documentation/src/main/antora/modules/ROOT/pages/_partials/generated_contributors.adoc
fdesiles, line 1: Why is this file modified?

undx: This file was generated during documentation update. It's usually generated at each builds.
As it follows all contributors' commits on project, it's always changing.

documentation/src/main/java/org/talend/runtime/documentation/Generator.java
undx: We don't generate changelog from old jira anymore (don't have ticket in mind) and it may break build.

PR #996, opened 2025-02-19
fix(QTDI-1188): fix unit test

ypiel-talend:
### Requirements

### DIFF
--- a/component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/service/jcache/FrontCacheResolverTest.java
+++ b/component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/service/jcache/FrontCacheResolverTest.java
@@ -56,7 +56,6 @@ class FrontCacheResolverTest {
 
     @Test
     void cleanupCaches() {
-        cacheResolver.cleanupCaches();
         assertEquals(0, cacheResolver.countActiveCaches());
         client.fetchIndex();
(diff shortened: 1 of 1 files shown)


component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/service/jcache/FrontCacheResolverTest.java
```diff
@@ -61,6 +61,9 @@ void cleanupCaches() {
     void clearCaches() throws JsonProcessingException {
         final int connectors = service.getConnectors().getPluginsList().size();
 
+        // First clean caches
+        cacheResolver.cleanupCaches();
```
ozhelezniak-talend, line 65: maybe in @BeforeEach method?

ozhelezniak-talend (approved):

component-server-parent/component-server/src/test/java/org/talend/sdk/component/server/service/jcache/FrontCacheResolverTest.java
ypiel-talend: Done

ozhelezniak-talend (approved):
LGTM

but I suppose now you can remove the first line about cleanupCaches from other tests too.

ozhelezniak-talend (approved):

ozhelezniak-talend:
Maybe chore?

PR #1000, opened 2025-02-27
Switching metawidget-core to load over HTTPS

coheigea:
### Requirements

### Why this PR is needed?

Fix for https://github.com/Talend/component-runtime/security/code-scanning/1

Nothing, just a switch from https to http to load a javascript resource


component-server-parent/component-server/src/dev/webapp/component.html
```diff
@@ -5,7 +5,7 @@ <h1>Sample</h1>
 <div id="metawidget">
 </div>
 <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
-<script src="http://metawidget.org/js/4.2/metawidget-core.min.js"></script>
+<script src="https://metawidget.org/js/4.2/metawidget-core.min.js"></script>
```
ypiel-talend, line 8: Should we set the sourceforge url since there is a 302:
```
$ curl https://metawidget.org/js/4.2/metawidget-core.min.js
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="https://metawidget.sourceforge.net/js/4.2/metawidget-core.min.js">here</a>.</p>
<hr>
<address>Apache/2.4.59 (Debian) Server at metawidget.org Port 443</address>
</body></html>
```

ypiel-talend (commented):
one comment

component-server-parent/component-server/src/dev/webapp/component.html
undx: Yep, but it fools security scanners apparently.

undx (approved):

ypiel-talend (approved):
LGTM

PR #1009, opened 2025-03-10
fix(QTDI-1345): [TCK]: tJDBCInput with dynamic type not works with the sql which have duplicated column name

wwang-talend:
### Requirements

### Why this PR is needed?
https://qlik-dev.atlassian.net/browse/QTDI-1345

https://qlik-dev.atlassian.net/browse/QTDI-1345


wwang-talend:
build fail not as current fix, seems inside .m2 issue

pyzhou-talend (approved):
LGTM

PR #1016, opened 2025-04-02
https://qlik-dev.atlassian.net/browse/QTDI-1219  Add documentation for LastGroup annotation

yyin-talend:
### Requirements

### DIFF
--- a/documentation/src/main/antora/modules/ROOT/pages/concept-processor-and-batch-processing.adoc
+++ b/documentation/src/main/antora/modules/ROOT/pages/concept-processor-and-batch-processing.adoc
@@ -88,6 +88,28 @@ public class BulkProcessor implements Serializable {
 }
 ----
 
+Some bulk api like SnowFlake, only can get which lines are rejected at last, not row level.
+
+It means we only can get reject info after every rows already go through elementListener method,
+so is impossible to output the reject records by elementListener method.
+Add a parameter in the AfterGroup method, use a new annotation “LastGroup” which is boolean,
+when “true” means that it is the last group can resolve this problem:
+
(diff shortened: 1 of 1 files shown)


documentation/src/main/antora/modules/ROOT/pages/concept-processor-and-batch-processing.adoc
```diff
@@ -88,6 +88,28 @@ public class BulkProcessor implements Serializable {
 }
 ----
 
+Some bulk api like SnowFlake, only can get which lines are rejected at last, not row level.
```
ypiel-talend, line 91: Please don't rely directly on a concrete use case (snowflake is too specific at first):

When using the @AfterGroup feature, it can be helpful to determine if it's the last call (last group). You can now annotate a Boolean with @LastGroup as a parameter of the method. This Boolean will be set to true for the final call, which can be useful for performing additional actions once all records have been processed.

For example, here is a scenario inspired by Snowflake:
- Batches of records are staged within each @AfterGroup call.
- Once all data has been staged, a final #commitBulkLoad call is executed. This action must be performed after all data has been staged.
- This process returns all rejected data, which is then transformed into TCK records and sent to the 'REJECT' flow.

```
@AfterGroup
public void afterGroup(@Output("REJECT") final OutputEmitter<Record> rejected, @LastGroup Boolean last) {
    database.pushToStaging(records);
    if(last){
        List<Data> dataToReject = database.commitBulkLoad();
        dataToReject.stream().map(d -> convertToRecord(d)).each(r -> rejected.emit(d));
    }
}
```

ypiel-talend (approved):
LGTM

PR #1017, opened 2025-04-07
feat(QTDI-1060): db mappings

undx:
**Issues**
- framework: https://qlik-dev.atlassian.net/browse/QTDI-1060
- studio: https://qlik-dev.atlassian.net/browse/QTUP-1717

**Related pull requests**
- https://github.com/Talend/component-runtime/pull/1017 (component-runtime)
- https://github.com/Talend/tdi-studio-se/pull/11197 (master)
- https://github.com/Talend/tcommon-studio-se/pull/8008 (master)

build pr
- https://github.com/Talend/tdi-studio-se/pull/11042 (maintenance/8.0.2)
- https://github.com/Talend/tcommon-studio-se/pull/7982 (maintenance/8.0.2)


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -1988,7 +1994,8 @@ private ComponentFamilyMeta getOrCreateComponent(final String component) {
                             iconFinder.findIcon(familyAnnotationElement), comp,
                             Class.class.isInstance(familyAnnotationElement)
                                     ? getPackage(Class.class.cast(familyAnnotationElement))
-                                    : ""))
+                                    : "",
```
ozhelezniak-talend, line 1997: same comment about the contstant NONE, where it's applicable

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -295,6 +295,12 @@ public String family() {
         public String[] categories() {
             return new String[] { "Misc" };
         }
+
+        @Override
+        public String databaseMapping() {
+            return "";
```
ozhelezniak-talend, line 301: here it's empty because it's NONE, or it means something different?

component-server-parent/component-server-model/src/main/java/org/talend/sdk/component/server/front/model/ComponentId.java
```diff
@@ -35,4 +35,6 @@ public class ComponentId {
     private String family;
 
     private String name;
+
+    private String databaseMapping;
```
ozhelezniak-talend, line 39: why do we call it so specific: "databaseMapping"?
Could it be typeMapping for more general use? (I'm searching for use-cases to understand what's behind it)

documentation/src/main/antora/modules/ROOT/pages/_partials/generated_actions.adoc
```diff
@@ -191,6 +191,17 @@ Mark a method as returning a Schema resulting from a connector configuration and
 ----
 
 
+= Schema Mapping
+
+Mark a method as returning a database mapping from a connector configuration and some other parameters.Use this annotation if database mapping can be dynamic and `@Components.databaseMapping` is set to `custom`.The functionality is for the Studio only.
```
ozhelezniak-talend, line 196: space between ".Use"

ozhelezniak-talend (dismissed):

sample-parent/sample-features/database-mapping/src/main/java/org/talend/sdk/component/sample/feature/databasemapping/service/UIService.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static final String SECOND_FLOW_NAME = "second";
+
+    public static final String THIRD_FLOW_NAME = "third";
+
+    @Service
+    private RecordBuilderFactory factory;
+
+    @DatabaseMapping
+    public String getDatabaseMapping(@Option("configuration") final Config configuration) {
```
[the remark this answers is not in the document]
undx: In studio when visiting the component and it has no static mapping (ie custom value).
However, for actions on tup side, we have a caching mecanism.

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
undx: not same parameter, it's packagename

ozhelezniak-talend (dismissed):
I'm worried that we put such specific action inside the core of Component.

component-server-parent/component-server-model/src/main/java/org/talend/sdk/component/server/front/model/ComponentId.java
undx: No, you should consider the studio's db mapping file like this one :
https://github.com/Talend/tcommon-studio-se/blob/9c89f86a1b094d789ebe068f7aa63f534[…]n/plugins/org.talend.core.runtime/mappings/mapping_Postgres.xml

This feature is in two parts, framework that allows to specify eventual db mapping according the studio mappings and the tup integration.
Giving the mapping allow to acccess the schema editors' db column type column .
spreading it via the SchemaInfo, the guess schema or directly in the editor.
Thus, this is mainly about schema editor access.

component-api/src/main/java/org/talend/sdk/component/api/component/DatabaseMapping.java
```diff
@@ -0,0 +1,57 @@
...
+     * The database type to map.
+     */
+    String value() default Mapping.NONE;
+
+    /**
+     * The mapper function name to fetch the custom mapping according configuration.
+     */
+    String mapping() default "";
+
+    public static class Mapping {
```
ozhelezniak-talend, line 46: Could you remind me, why we avoid using enums?

ozhelezniak-talend:
The implementation is still too specific to the studio. It uses Studio's implementation of file mapping as a core element. 
During the meetings that I've heard it was introduced differently. 
Clearly some miscommunication. 
Maybe QLIK cloud can use this RAW DB type thing as well, so we might need to have some kind of mapping there too. It's just a thought.
@ypiel-talend what do you think (as proxy-po)

component-runtime-manager/src/test/java/org/talend/sdk/component/runtime/manager/extension/ComponentSchemaEnricherTest.java
```diff
@@ -57,6 +60,33 @@ void fixedSchemaMetadataNotPresent() {
...
+    }
+
+    @Test
+    void mysqlMappingMetadataPresent() {
+        assertEquals(new HashMap<String, String>() {
+
+            {
+                put("tcomp::ui::schema::mapping", "mysql");
+            }
+        }, enricher.onComponent(ProcessorWithMySqlMapping.class, ProcessorWithMySqlMapping.class.getAnnotations()));
```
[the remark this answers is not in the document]
undx: for now, using java 8 compilation level. not relevant

ozhelezniak-talend (approved):
Contacted Yves.
Looks like the implementation approach was agreed.
It's studio specific.
Maybe if we find smth better in future, we would be able to deprecate this one and replace it with a new one.
Conceptually this project becomes less and less public and universal =)

PR #1019, opened 2025-04-08
feat(QTID-675): Some changes in checkpoint sample.

ypiel-talend:
### Requirements


sample-parent/sample-features/checkpoint-runner/src/main/java/org/talend/sdk/component/sample/feature/checkpoint/CheckpointInput.java
```diff
@@ -179,10 +174,6 @@ public static class CheckPointConfig implements Serializable {
         @Option
         @Documentation("Checkpointing state : since id.")
         private int sinceId;
-
-        @Option
-        @Documentation("Checkpointing state : status.")
-        private String status;
```
undx, line 185: If status is removed, need to update the `README.md`

undx (approved):

PR #1027, opened 2025-04-14
fix(QTDI-1466): Fix chained input

ypiel-talend:
https://qlik-dev.atlassian.net/browse/QTDI-1466

### Requirements

### DIFF
--- a/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/ChainedInput.java
+++ b/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/ChainedInput.java
@@ -15,6 +15,7 @@
  */
 package org.talend.sdk.component.runtime.manager.chain;
 
+import java.util.Optional;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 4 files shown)


component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/chain/ChainedInput.java
```diff
@@ -88,7 +93,15 @@ public void stop() {
         }
     }
 
-    private void initDelegate() {
-        delegate = parent.getIterator().hasNext() ? parent.getIterator().next().create() : null;
+    private Input initDelegate() {
```
ozhelezniak-talend, line 96: with this return type it's more: "pickDelegate"
because you don't set a field value here anymore and nothing is initialized inside the method
we can say that it "startDelegate" cause it's executed the method

ozhelezniak-talend (approved):
looks ok
I just don't like current name "initDelegate" I think it doesn't explain its functionality anymore

ozhelezniak-talend (approved):

PR #1034, opened 2025-04-22
fix(QTDI-1393): move to gradle jetbrains plugin

ozhelezniak-talend:
https://qlik-dev.atlassian.net/browse/QTDI-1393


talend-component-kit-intellij-plugin/.gitignore
```diff
@@ -0,0 +1,5 @@
+.gradle
+.idea/
```
undx, line 2: I think that some of those entries are already in top level .gitignore.

talend-component-kit-intellij-plugin/gradle/wrapper/gradle-wrapper.jar
undx, line 1: Maybe we could install gradle like we do for node and npm...

talend-component-kit-intellij-plugin/.gitignore
undx, line 2: I think that some of those entries are already in top level .gitignore.

undx (dismissed):
lgtm 
minor comments

talend-component-kit-intellij-plugin/.gitignore
ozhelezniak-talend: I was just to make the developers life easier. 
I imported that folder as a separate project and my Idea didn't see some .gitignore exceptions from the parent root. 
(for example it didn't exclude "build" folder)

talend-component-kit-intellij-plugin/gradle/wrapper/gradle-wrapper.jar
ozhelezniak-talend: I saw that for most of the time people include those into the file into git as well
(I was surprised, but I don't remember the reason, maybe for github actions...)

ozhelezniak-talend: If you want I can try to change it a bit
Or we can do it with the next PR...

undx (approved):

PR #1038, opened 2025-04-30
chore(QTDI-1393): Create build and release scripts for intelliJ TCK plugin

acatoire:
Implement build and release script based on Jetbrains recommandation.
Part of automation task for https://qlik-dev.atlassian.net/browse/QTDI-1516


talend-component-kit-intellij-plugin/scripts/build.sh
```diff
@@ -0,0 +1,54 @@
...
+
+echo "==> Running tests"
+cd ../..
+$GRADLE_CMD check || {
+  echo "==> Tests failed - collecting results"
+  cp -r ./build/reports/tests ./test-results
+}
+
+echo "==> Verifying plugin with IntelliJ Plugin Verifier"
+mkdir -p "$PLUGIN_VERIFIER_HOME/ides"
```
ozhelezniak-talend, line 51: this DIR should be on EFS, it will contain and IDEA installations that can be re-used between builds, and they should be quite big, and it takes time to load

talend-component-kit-intellij-plugin/scripts/build.sh
```diff
@@ -0,0 +1,54 @@
...
+$GRADLE_CMD verifyPlugin -Dplugin.verifier.home.dir="$PLUGIN_VERIFIER_HOME"
```
ozhelezniak-talend, line 52: also, verify we can make optional. 
Like "Run Tests" it takes time, and sometimes we may not want to verify (but this improvement is optional)

EFS for ides - I think "have to" to manage it from the begining

ozhelezniak-talend (dismissed):
LGTM

ozhelezniak-talend (approved):

PR #1063, opened 2025-06-13
fix(QTDI-1656): what is a processor

thboileau:
### Requirements

### DIFF
--- a/documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
+++ b/documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
@@ -76,7 +76,7 @@ The component type is declared in the component class. When you import your proj
 
 *Processor*/*Output* components can be:
 
-- A processor. `@Processor` is the default for output components. A method decorated with `@Processor` is considered as a producer factory.
+- A processor. `@Processor` is the default for output components.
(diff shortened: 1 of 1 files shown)


documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
```diff
@@ -76,7 +76,7 @@ The component type is declared in the component class. When you import your proj
 
 *Processor*/*Output* components can be:
```
thboileau, line 77: to be honest, this is not clear at all.
I expect an answer to the following question: what are the characteristics of a Processor. What's the contract?
And I see two things: a class with a `@Processor` annotation and a method annotated `@ElementListener` with an input record and and output record.

Then, what's the purpose to mention `Combiner` which is an answer to another question: What is not a processor?

Then it is not clear what are the characteristics of an output? There is just no answer to this question. If we want to say that a processor can serve as an output, we must add what are the differences.

Another point, the documentation does not mention clearly another aspect of components: how to link them.
From my current understanding, there are two kind of links:

- records that can be generated by an input component, a processor. the second component is "activated" when a record is generated, and it should be either a processor or an output component
- "subjob" links (OK or KO): the second component is activated when the first one is over. second component could whatever that can start without record: input or standalone components.

PR #1074, opened 2025-07-22
fix(QTDI-1814): fix vulns in npm

undx:
https://qlik-dev.atlassian.net/browse/QTDI-1814


yyin-talend:
Duplicated/fix in QTDI-1403

PR #1076, opened 2025-07-22
fix(QTDI-1815): fix component-tools npm vulns

undx:
### Requirements


yyin-talend:
Duplicated/fix in QTDI-1403

PR #1086, opened 2025-08-05
feat(QTDI-1672): Allow disable nullable check in tck framework level

yyin-talend:
https://qlik-dev.atlassian.net/browse/QTDI-1672
### Requirements

### DIFF
--- a/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
+++ b/component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
@@ -327,7 +327,8 @@ public final class RecordImpl implements Record {
                         "Entry '" + entry.getOriginalFieldName() + "' expected to be a " + entry.getType() + ", got a "
                                 + type);
             }
-            if (value == null && !entry.isNullable()) {
+            final boolean disableNullableCheck = Boolean.parseBoolean(System.getProperty("talend.sdk.nullable.check", "false"));
+            if (!disableNullableCheck && value == null && !entry.isNullable()) {
                 throw new IllegalArgumentException("Entry '" + entry.getOriginalFieldName() + "' is not nullable");
             }
             return entry;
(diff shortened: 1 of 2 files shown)


component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -587,11 +591,15 @@ private <T> Builder append(final Schema.Entry entry, final T value) {
...
-                throw new IllegalArgumentException(realEntry.getName() + " is not nullable but got a null value");
-            }
+//            if (Boolean.parseBoolean(System.getProperty("talend.sdk.skip.nullable.check", "false"))) {
+//                values.put(realEntry.getName(), value);
+//            } else {
+                if (value != null) {
+                    values.put(realEntry.getName(), value);
+                } else if (!realEntry.isNullable()) {
+                    throw new IllegalArgumentException(realEntry.getName() + " is not nullable but got a null value");
+                }
```
undx, line 601: same code as before, no check is done.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
```diff
@@ -357,7 +359,9 @@ public Record build() {
                         .filter(it -> !it.isNullable() && !values.containsKey(it.getName()))
                         .map(Schema.Entry::getName)
                         .collect(joining(", "));
-                if (!missing.isEmpty()) {
+
+                if (!Boolean.parseBoolean(System.getProperty("talend.sdk.skip.nullable.check", "false"))
```
undx, line 363: see main review comment.

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/record/RecordImpl.java
yyin-talend: For this issue, only need to add the check in build(), but not in validateTypeAgainstProvidedSchema, discussed with wwang and confirmed from him. 
It is used only for build, for validation it is still needed.

PR #1105, opened 2025-09-22
fix(QTDI-2004): check if running into jobserver

undx:
https://qlik-dev.atlassian.net/browse/QTDI-2004

### DIFF
--- a/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
+++ b/component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
@@ -15,6 +15,7 @@
  */
 package org.talend.sdk.component.runtime.manager;
 
+import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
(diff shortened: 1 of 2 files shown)


ozhelezniak-talend (approved):

component-runtime-manager/src/main/java/org/talend/sdk/component/runtime/manager/ComponentManager.java
```diff
@@ -213,11 +214,14 @@ private static ComponentManager buildNewComponentManager() {
...
-                    parallelIf(Boolean.getBoolean("talend.component.manager.plugins.parallel"),
-                            container.getDefinedNestedPlugin().stream().filter(p -> !hasPlugin(p)))
-                            .forEach(this::addPlugin);
-                    info("Components: " + availablePlugins());
+                    try {
+                        parallelIf(Boolean.getBoolean("talend.component.manager.plugins.parallel"),
+                                container.getDefinedNestedPlugin().stream().filter(p -> !hasPlugin(p)))
+                                .forEach(this::addPlugin);
+                        info("Components: " + availablePlugins());
+                    } catch (Exception e) {
```
wwang-talend, line 222: seems here only catch the exception and ignore it, is enough?

PR #1188, opened 2026-04-07
doc(QTDI-1656): misc documentation issues

thboileau:
### Requirements

### Why this PR is needed?
Fix miscellaneous issues reported in ticket https://qlik-dev.atlassian.net/browse/QTDI-1656

### DIFF
--- a/documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
+++ b/documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
@@ -76,7 +76,7 @@ The component type is declared in the component class. When you import your proj
 
 *Processor*/*Output* components can be:
 
-- A processor. `@Processor` is the default for output components. A method decorated with `@Processor` is considered as a producer factory.
+- A processor. `@Processor` is the default for output components.
 - Combiners are not supported by the framework. Combiners allow to aggregate results in a single partition.
 
 *Standalone* component can be only a driver runner.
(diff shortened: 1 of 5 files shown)


documentation/src/main/antora/modules/ROOT/pages/component-registering.adoc
```diff
@@ -76,7 +76,7 @@ The component type is declared in the component class. When you import your proj
 
 *Processor*/*Output* components can be:
 
-- A processor. `@Processor` is the default for output components. A method decorated with `@Processor` is considered as a producer factory.
+- A processor. `@Processor` is the default for output components.
```
thboileau, line 79: to be honest, this is not clear at all. and the whole chapter could be rephrased.

I expect an answer to the following question: what are the characteristics of a Processor? What's the contract?
And I see two characteristics:

a class with a @Processor annotation that allows to name the processor.
a mandatory method annotated @ElementListener with an input record and and output record.
Then, what's the purpose to mention Combiner which is an answer to another question: What is not a processor or an output?

Then it is not clear what are the characteristics of an output? There is just no answer to this question. If we want to say that a processor can serve as an output, we must add what are the differences between processor and output.

Another point related to the previous comment, the documentation does not mention clearly another aspect of components: how to link them.
From my current understanding, there are two kind of links:

records that can be generated by an input component, a processor. the second component is "activated" when a record is generated, and it should be either a processor or an output component
"subjob" links (OK or KO): the second component is activated when the first one is over. second component could whatever that can start without record: input or standalone components.

dmytro-sylaiev (approved):

PR #1251, opened 2026-07-09
doc(QTDI-3156): Improve docs for CreateConnection service

dmytro-sylaiev:
### Requirements

### Why this PR is needed?

* Missing info in documentation about CreateConnection service behavior at studio-integration plugin

* Doc change only

### AI generated code
https://internal.qlik.dev/general/ways-of-working/code-reviews/#guidelines-for-ai-generated-code
- [] this PR has been written with the help of GitHub Copilot or another generative AI tool

### DIFF
--- a/component-api/src/main/java/org/talend/sdk/component/api/service/connection/CreateConnection.java
+++ b/component-api/src/main/java/org/talend/sdk/component/api/service/connection/CreateConnection.java
@@ -28,8 +28,8 @@ import org.talend.sdk.component.api.service.ActionType;
 @Target(METHOD)
 @Retention(RUNTIME)
 @Documentation("Mark an action works for creating runtime connection, returning a runtime connection object like jdbc connection if database family. "
-        + "Its parameter MUST be a datastore. Datastore is configuration type annotated with @DataStore. "
-        + "The `value` parameter must match the runtime class constructor parameter name of the component configuration class to correctly apply `@ActiveIf` in the Studio UI. "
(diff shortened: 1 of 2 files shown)


component-api/src/main/java/org/talend/sdk/component/api/service/connection/CreateConnection.java
```diff
@@ -29,6 +29,7 @@
 @Retention(RUNTIME)
 @Documentation("Mark an action works for creating runtime connection, returning a runtime connection object like jdbc connection if database family. "
         + "Its parameter MUST be a datastore. Datastore is configuration type annotated with @DataStore. "
+        + "The `value` parameter must match the runtime class constructor parameter name of the component configuration class to correctly apply `@ActiveIf` in the Studio UI. "
```
thboileau: I'm afraid that in this this context there could be a confusion between:

1. `value` parameter of the `@CreateConnection` annotation
2. `value` parameter of the `@Option` annotation of the method annotated `@CreateConnection`

Is something like this clearer?
```
Mark a method as returning a runtime connection object like jdbc connection if database family. 
The method's parameter MUST be a datastore annotated `@Option`.
The `value` parameter of `@Option` must match the runtime class constructor parameter name of the component configuration class to correctly apply `@ActiveIf` in the Studio UI. 
```

thboileau (dismissed):

thboileau (approved):

PR #1253, opened 2026-07-13
fix(QTDI-2709): TCK processors should send generated records to next connector without buffer

wwang-talend:
### Requirements

### Why this PR is needed?
https://qlik-dev.atlassian.net/browse/QTDI-2709

https://qlik-dev.atlassian.net/browse/QTDI-2709

### AI generated code
https://internal.qlik.dev/general/ways-of-working/code-reviews/#guidelines-for-ai-generated-code
- [x] this PR has been written with the help of GitHub Copilot or another generative AI tool


wwang-talend (commented):
## Code Review — PR #1253

### 🔴 Critical: `ClassCastException` in non-DI contexts

`ProcessorImpl.buildProcessParamBuilder()` and `toOutputParamBuilder()` cast `outputs.create(name)` to `OutputIterator`:

```java
// ProcessorImpl.java:156
return (inputs, outputs) -> (OutputIterator) outputs.create(name);
```

But `OutputFactory.create()` returns `OutputEmitter`. Only the DI/Studio `OutputsHandler` returns objects implementing both interfaces. **Two other `OutputFactory` implementations do NOT implement `OutputIterator`:**

- `BeamOutputFactory` / `BeamOutputEmitter` (`BaseProcessorFn.java:162-176`)
- `DataOutputFactory` / `OutputEmitterImpl` (`JobImpl.java:661-681`)

Any processor using `@Output OutputIterator<Record>` in a Beam runner or via `JobImpl` chain will throw `ClassCastException` at runtime.

**Suggested fix:** Either update `OutputFactory` to provide `OutputIterator` support (e.g., a default method), or have `ProcessorImpl` wrap the `OutputEmitter` in an adapter, or update all `OutputFactory` implementations.

---

### 🟡 Design: Implicit contract not enforced by type system

The `OutputFactory` interface declares `OutputEmitter create(String name)`. This PR adds an implicit requirement that returned objects also implement `OutputIterator`, but nothing enforces this. Future `OutputFactory` implementations will hit the same `ClassCastException` silently at runtime.

---

### 🟡 Raw types throughout

`OutputEmitterWithIterator` implements raw `OutputEmitter` and `OutputIterator` (no type parameters). `setIterator()` takes raw `Iterator`. This bypasses generic type safety and produces compiler warnings.

---

### 🟡 Silent exception swallowing

`BaseIOHandler.IO.closeSource()` catches all exceptions with an empty block:
```java
} catch (final Exception e) {
    // best effort cleanup
}
```
Consider at minimum logging at DEBUG/TRACE level to avoid hiding bugs during iterator cleanup.

---

### 🟡 Test reliability concerns

- Memory-based assertions using `System.gc()` + `Runtime.freeMemory()` are non-deterministic. The 5MB threshold could be flaky across JVMs, GC algorithms, or CI environments.
- `builderFactory` is a `protected static` mutable field — risky with parallel test execution.

wwang-talend:
OutputIterator only for studio, only studio have that case, only need to document it that not suitable for beam env.

wwang-talend (commented):
## Response to AI Review

### 🔴 ClassCastException in non-DI contexts

Valid observation. However, this is **by design** — the `OutputIterator` pattern was built specifically for the Studio DI runtime to solve the single-threaded buffering problem.

As stated in the spec: *"The impact of this pattern on Cloud is minimal. In the Cloud, processors are not TCK processors (usually dataprep functions)."*

No real-world processor will use `@Output OutputIterator<T>` in a Beam pipeline. Adding adapter/fallback code in `BeamOutputFactory` / `JobImpl` would add unnecessary complexity for a use case that does not exist.

**Action taken:** Added explicit javadoc on `OutputIterator` documenting that it is **Studio DI only** and not supported in Beam runners.

---

### 🟡 Implicit contract not enforced by type system

Accepted trade-off. We control all `OutputFactory` implementations. The `ModelVisitor` validates processor declarations at plugin load time, so misuse surfaces early. A `default` method on `OutputFactory` would be cleaner but is a bigger change with no practical benefit given the Studio-only scope.

---

### 🟡 Raw types

Fair point. Will address in a follow-up cleanup.

---

### 🟡 Silent exception swallowing

Agreed. Will add logging in the catch block.

---

### 🟡 Test reliability

The memory test is intentionally designed as a manual/observable test — it prints memory stats so developers can see the delta. The 5MB threshold is generous (actual delta is ~1-2MB with iterator mode). The `static builderFactory` pattern is consistent with existing tests (`DIBulkAutoChunkTest`, `DIBatchSimulationTest`).

wwang-talend:
### Clarification on `DataOutputFactory` / `JobImpl`

The reviewer flagged `DataOutputFactory` (`JobImpl.java:661-681`) as another non-DI `OutputFactory` that would throw `ClassCastException`.

**Analysis:** `JobImpl` with `DataOutputFactory` is the **local chain execution** path used by the `Job.builder()` programmatic API (`component-runtime-manager`). It is NOT Studio DI and NOT Beam.

However, this API is only referenced in documentation examples — no production usage exists. The real Studio DI runtime uses `OutputsHandler` which correctly implements both `OutputEmitter` and `OutputIterator`.

The `OutputIterator` javadoc already clarifies it is supported in Studio DI runtime only. This is sufficient — adding adapter code in `JobImpl.DataOutputFactory` would be unnecessary complexity for a path that will not encounter `OutputIterator` processors in practice.

wwang-talend:
sonar often report some small smell in tests, not worth to work on it even, should care on the major things we need to care thing, not that smell even in junit tests only.

Every thing good smell, but bug appear. That is a kind of context go out like AI.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
```diff
@@ -70,4 +65,53 @@ public OutputFactory asOutputFactoryForGuessSchema() {
         };
     }
 
+    private Object convert(final Object value, final BaseIOHandler.IO ref) {
+        if (value == null) {
+            return null;
+        } else if (value instanceof javax.json.JsonValue) {
+            return jsonb.fromJson(value.toString(), ref.getType());
+        } else if (value instanceof Record record) {
```
thboileau: use `rec` for instance ?

wwang-talend: done

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/output/ProcessorImpl.java
```diff
@@ -150,10 +151,11 @@ public void beforeGroup() {
 
     private BiFunction<InputFactory, OutputFactory, Object> buildProcessParamBuilder(final Parameter parameter) {
         if (parameter.isAnnotationPresent(Output.class)) {
-            return (inputs, outputs) -> {
-                final String name = parameter.getAnnotation(Output.class).value();
-                return outputs.create(name);
-            };
+            final String name = parameter.getAnnotation(Output.class).value();
+            if (OutputIterator.class == parameter.getType()) {
+                return (inputs, outputs) -> (OutputIterator) outputs.create(name);
```
thboileau: That only works because the one concrete implementation, `OutputsHandler.OutputEmitterWithIterator`, happens to implement both interfaces. Nothing in OutputFactory's signature says a factory may hand back something iterator-capable.

One solution is to make the capability explicit in the `OutputFactory` : a default `OutputIterator createIterator(String name)` that throws a clear UnsupportedOperationException by default, except in the case of the `OutputsHandler` class.


The default implementation of `createIterator` allows retro-compatibility
```java
public interface OutputFactory {

    OutputEmitter create(String name);

    default OutputIterator createIterator(String name) {
        throw new UnsupportedOperationException();
    }
}
```

In `OutputsHandler`:
```java
    public OutputFactory asOutputFactory() {
        return new OutputFactory() {
            @Override
            public OutputEmitter create(final String name) {
                final BaseIOHandler.IO ref = connections.get(getActualName(name));
                return value -> {
                    if (ref != null && value != null) {
                        ref.add(convert(value, ref));
                    }
                };
            }

            @Override
            public OutputIterator createIterator(final String name) {
                final BaseIOHandler.IO ref = connections.get(getActualName(name));
                return iterator -> {
                    if (ref == null) {
                        return;
                    }
                    ref.setSource(new Iterator() {

                        @Override
                        public boolean hasNext() {
                            return iterator.hasNext();
                        }

                        @Override
                        public Object next() {
                            return convert(iterator.next(), ref);
                        }
                    });
                };
            }
        };
    }
```


What do you think about this proposal?

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
```diff
@@ -91,19 +91,32 @@ static class IO<T> {
 
         private final Class<T> type;
 
+        private Iterator<T> source;
```
thboileau, line 244: I feel we need more documentation here.

First, the `IO` class is shared between the `InputsHandler` and the `OutputsHandler` classes, and implements a "push-or-pull" behavior which is only used by the `OutputsHandler`. People may read InputsHandler and wonder why "source" is never used. We should document it.

Secondly, as we know that "push-or-pull" is actually exclusive, we could think of a better design that clearly separates these two behaviors, but I don't how, I'm not smart enough. When IO is created, there is no notion of "push" or "pull", just the name of the connection. So documentation would help here

What do you think about this?

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/output/ProcessorImpl.java
wwang-talend: Hi @thboileau , thanks for the review.

OutputFactory is a public interface, but not so public as only studio common javajet use it. So from the history regression risk for future view, i think your idea is ok, which add a new interface method.

My origin idea is not change interface, keep complex in implement.

So i will consider your advice as it look good, but give me some times to prove your idea is real better.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
wwang-talend: about refactor, no code compatiable risk for future as that is not so public to client user, we can do later. So agree document enough now.

wwang-talend: done

component-runtime-impl/src/main/java/org/talend/sdk/component/runtime/output/ProcessorImpl.java
wwang-talend: done

thboileau (dismissed):

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
```diff
@@ -114,6 +142,16 @@ void add(final T e) {
         Class<T> getType() {
             return type;
         }
+
+        private void closeSource() {
```
wwang-talend, line 282: the autoclose not work yet, except we call that .reset() method some where in studio common javajet, but in fact, not. So i may remove this part.

wwang-talend: for httpclient==>
every input record==> make M(bigger enough to use streaming feature here) output records

we may need to use this autoclose to release the resource for pre streaming object asap? not wait on @PreDestory close method.

But studio allow user not draw output line even allow. So for safe, better to close in @PreDestory too. Above one for "close asap", @PreDestory for "close finally even outside exception happen and so on"

wwang-talend: so keep it, not public and not make code refactor complex in future.

wwang-talend:
about parallize, i not find issue for currernt design for studio part. But need to test

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/schema/TaCoKitGuessSchema.java
```diff
@@ -776,7 +776,8 @@ public void fromOutputEmitterPojo(final Processor processor, final String outBra
                     .flatMap(m -> IntStream
                             .range(0, m.getParameterCount())
                             .filter(i -> m.getParameters()[i].isAnnotationPresent(Output.class)
-                                    && outBranchName.equals(m.getParameters()[i].getAnnotation(Output.class).value()))
+                                    && java.util.Arrays.asList(m.getParameters()[i].getAnnotation(Output.class).value())
```
thboileau, line 780: can the Output annotation contains several "values"?

Objects.equals(m.getParameters()[i].getAnnotation(Output.class).value(), outBranchName) should be enough?

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
```diff
@@ -69,41 +96,178 @@ public void addConnection(final String connectorName, final Class<?> type) {
...
+        }
         return (T) connections.get(connectorName).next();
     }
 
     public boolean hasMoreData() {
-        return connections.entrySet().stream().anyMatch(e -> e.getValue().hasNext());
+        if (taggedSource != null) {
+            return pendingOutputName != null || taggedSource.advance();
+        }
+        for (IO value : connections.values()) {
```
thboileau, line 124: I feel that this loop is exactly

`connections.entrySet().stream().anyMatch(e -> e.getValue().hasNext());`

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
```diff
@@ -0,0 +1,117 @@
...
+ *     out.setIterator("REJECT", errorLog.lazyRead());
+ * }
+ * }
+ * </pre>
+ *
+ * <p>
+ * The drain loop on the Studio side uses {@code hasDataFor(name)} for correct
+ * per-connection checking in both modes:
+ *
+ * <pre>
```
thboileau: I think that this part does not relate directly to the MultiIterator.
It's part of the API of BaseIOHandler class, can be implemented based on the current state of BaseIOHandler (ie.: with only the support of OutputEmitter). It's a nice replacement of the `if (row2 != null) {` that we can find in the generated code.

```
						while (outputHandler_tDateFormatValidatorValidator_1.hasMoreData()) {
							nbLineOutput_tDateFormatValidatorValidator_1++;
							row2 = outputHandler_tDateFormatValidatorValidator_1.getValue("FLOW");
							row3 = outputHandler_tDateFormatValidatorValidator_1.getValue("reject");

							/**
							 * [tDateFormatValidatorValidator_1 process_data_begin ] stop
							 */

// Start of branch "row2"
							if (row2 != null) {

								/**
								 * [tLogRow_1 main ] start
								 */
```

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
```diff
@@ -33,16 +37,57 @@ public OutputsHandler(final Jsonb jsonb, final Map<Class<?>, Object> servicesMap
...
+            public <T> MultiOutputIterator<T> createMultiOutputIterator() {
+                return new MultiOutputIterator<T>() {
+
+                    @Override
+                    public void setIterator(final Iterator<TaggedOutput<T>> iterator) {
+                        setTaggedSource(() -> {
+                            if (!iterator.hasNext()) {
+                                return false;
+                            }
+                            final TaggedOutput<T> tagged = iterator.next();
```
thboileau, line 62: I feel that the result of the conversion could put in the right IO.
By doing so, I feel there is no need of taggedSource, pendingRecord and pendingOutputName in class BaseIOHandler

What do you think?

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
```diff
@@ -0,0 +1,117 @@
...
+     * Call once per output that needs a dedicated lazy source; connections without an
+     * assigned iterator fall back to the push-mode queue as usual.
+     *
+     * <p>
+     * Mutually exclusive with {@link #setIterator(Iterator)} within one invocation.
+     *
+     * @param outputName the output connection name (e.g. {@code "MAIN"}, {@code "REJECT"})
+     * @param iterator the lazy iterator producing records for that connection
+     */
+    void setIterator(String outputName, Iterator<T> iterator);
```
thboileau, line 101: I really wonder if we should drop this in another interface. It's another user case.

wwang-talend: that for the case which no need use TaggedOutput, like httpclient. And also sure httpclient can use TaggedOutput.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
wwang-talend: yes, can. And i also let AI do it before, but no stable result, after AI refactor, the code more hard to read and not better. So keep this version.

I will try again, but not promise that as the code is clean now.

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
wwang-talend: that is used here : https://github.com/Talend/studio/pull/1672/changes#diff-803bb6ddc517da716b80b39a1dffd50b4fc5b3fe14b4c65c3dbca953bcc754f6L35

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
thboileau: yes, that's why I prefer another interface, like SingleOutputIterator for instance

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
```diff
@@ -33,16 +37,57 @@ public OutputsHandler(final Jsonb jsonb, final Map<Class<?>, Object> servicesMap
...
+                            final String name = tagged.getOutputName();
+                            final BaseIOHandler.IO ref = connections.get(name);
```
thboileau, line 64:
```suggestion
                            final BaseIOHandler.IO ref = connections.get(getActualName(name));
```

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
wwang-talend: about this code generation, the patterns : 

pattern 1: one input record come ==> dispatch output record, go output 1(main) or go output 2(reject) or output xxx, but only go unique output x.

pattern 2: any, group or not group, or one by one ==> output x records to output 1(main), output y to output 2(reject), z to output 3...,x, y, z value even no relationship with input records number

About pattern 1, the code generation above is right.

About pattern 2, the code generation above is strange. 

TODO desc more

thboileau: What I want to say is that documenting outputsHandler.HasMoreData is right in class BaseIOHandler, not MultiOutputIterator.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
wwang-talend: stream() have performance cost, this is a high called methods, here just improve performance. But maybe i over consider?

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
wwang-talend: i remove getActualName on purpose, as no meaning, and a equals method in a method called often.

wwang-talend: that just convert "__default__" to "FLOW", connector developer never use "__default__", that is a hiden thing in connector runtime.

wwang-talend: but maybe over consider performance

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
wwang-talend: image one day:
no input==>httpclient ==> main
                ==> reject
                ==> xxxx

a number input==>httpclient ==> x number main
                ==> y reject
                ==> z xxxx
a value no relationship with x + y + z and so on.

MultiOutputIterator support one channel too. But SingleOutputIterator don't support multi channel.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
thboileau: let's forget this. This is a minor point.

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
wwang-talend: aha, i will correct the doc, thanks. TODO

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
thboileau: yes, I understand. to me these are two different uses case that could be handled by tho distinct interfaces.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
wwang-talend: i revert it as "Record record" is right expression, here i not agree sonar.

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/schema/TaCoKitGuessSchema.java
wwang-talend: yes, can define like this now:
```
@Output({"MAIN", "REJECT", "XXX"}) MultiOutputIterator<Record> output
```

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
thboileau: nb: let's start with your solution. I propose to discuss the topic with Yves and Manu when they will be back.

component-api/src/main/java/org/talend/sdk/component/api/processor/MultiOutputIterator.java
wwang-talend: doc fixed

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/schema/TaCoKitGuessSchema.java
thboileau: ok

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
thboileau: ok

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/OutputsHandler.java
thboileau: in this case, shouldn't we drop the other references to `getActualName` ? Either we keep getActualName or we drop it. we can't be in the middle because it makes maintenance difficult.

wwang-talend: i just remove it for the new interface, if want to remove it for old interfact, i need more deep check, and that is not about current jira, we can do it in future.

thboileau (approved):
ok for now. will need to discuss the current solution with Yves and Emmanuel.

wwang-talend:
have review all sonar comments

component-studio/component-runtime-di/src/main/java/org/talend/sdk/component/runtime/di/BaseIOHandler.java
```diff
@@ -114,6 +278,16 @@ void add(final T e) {
         Class<T> getType() {
             return type;
         }
+
+        private void closeSource() {
+            if (source instanceof AutoCloseable) {
+                try {
+                    ((AutoCloseable) source).close();
```
undx, line 285: comestic and sonar friendly proposition
```suggestion
            if (source instanceof AutoCloseable closeable) {
                try {
                    closeable.close();
```

undx (approved):
lgtm 👍🏻

